PR #24, opened 2014-11-05
make fit_series work with hdf5 video files

tdimiduk:
This patch set updates holopy's timeseries to be able to work with the hdf5 files the camera controller now generates. 

It also adds some preliminary work on fitting to hdf5 timeseries and writing hdf5 fit output. That code is working, but probably still a little unstable, so maybe don't depend on it. 

This is kind of a dirty set of code that I am merging because it contains stuff that is useful, and I am not sure when I will get a chance to finish it. It keeps around fit_series which works on directories of tiffs, and adds series.fit which works on hdf5 files. Eventually these two should probably be unified in some manner.

### DIFF
--- a/holopy/core/errors.py
+++ b/holopy/core/errors.py
@@ -24,41 +24,13 @@ Error classes used in holopy
 from __future__ import division
 
 
-
-class Error(Exception):
-    def __init__(self, message = ""):
-        self.message = message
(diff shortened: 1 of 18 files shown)


barkls:
upcoming implementation of xarray will solve this.

PR #36, opened 2016-06-07
centerfinder blurring

barkls:
Added a blur to centerfinder before Hough transformto help find the correct hologram centre in the presence of localized regions of large gradients, such as dust particles. This required updating the expected centerfinder test results as well.


tdimiduk:
You should probably not do the blurring if blursize = 0 (put in an if check). That will allow us to recover the old code behaviour completely, and make the code run slightly faster in that case

PR #62, opened 2016-08-30
Added some tests of inference code

barkls:
These tests aren't meant to be comprehensive. mcmc.py was also modified to allow passing of random seeds to Emcee and correctly handle non-parallel computation. 

I'm not sure why there are 3 commits showing up. As far as I know two of them are old and already merged. It's only the most recent one that is relevant.


tdimiduk:
The seed management may still need improvement (if you call Emcee.make_sampler multiple times it will use the same seed), however, I think we are going to rework that code enough that fixing that should more be seen as a todo for the rework of the mcmc code rather than something that needs to happen with this pr.

PR #71, opened 2016-09-13
Allow user to not specify polarization vector for single sphere scatterers

barkls:
Tom: calc_holo thought it could run without optics, since calc_field was running interpret_args. However, calc_holo then needs to pass optics to scattered_field_to_hologram. I added interpret_args directly into calc_holo and calc_intensity. It would be better if calc_field passed its arguments up since we are now running interpret_args twice, but I couldn't figure out a good way to do this. Alternatively, we could move the redundant interpret_args from calc_holo into scattered_field_to_hologram but I don't see an advantage here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ holopy/propagation/tests/gold/full_data
 holopy/scattering/tests/gold/full_data
 .coverage
 .nfs*
+*~
--- a/holopy/core/marray.py
+++ b/holopy/core/marray.py
@@ -183,12 +183,6 @@ class Schema(HoloPyObject):
     def positions(self, val):
         self._positions = val
 
-    # TODO: put this somewhere sensible, make it handle phi as well
-    def positions_theta_phi(self):
-        if isinstance(self.positions, Angles):
-            return self.positions.positions_theta_phi()
-        else:
-            raise UnspecifiedPosition()
 
     # TODO: need equivalents for set_metadata or to rewrite other code so it
     # doesn't need it.
--- a/holopy/core/metadata.py
+++ b/holopy/core/metadata.py
@@ -178,6 +178,9 @@ class Positions(np.ndarray, HoloPyObject):
 
         return np.vstack((r, theta, phi)).T
 
+    def theta_phi(self, origin):
+        return self.r_theta_phi(origin)[:, 1:]
+
     def kr_theta_phi(self, origin, wavevec):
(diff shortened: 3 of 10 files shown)


tdimiduk:
You should test (in a unit test) that calc_field returns the same fields for a sphere when given polarization = (1, 0) and polarization = (0, 1). Then document that the two things are equivalent (pointing at the unit test). That way people like me who are only 90% confident that it is fine to just pick a polarization for a symmetric scatterer don't have to worry. 

It doesn't look like you need to separately call interpret_args in calc_intensity and calc_holo, since they call calc_field and calls interpret_args with exactly the same arguments.

barkls:
That unit test is a good idea and I will implement it. 

calc_field does call interpret_args, but it can't pass the {theory, optics, locations} back up to calc_holo. When calc_holo then tries to run scattered_field_to_hologram, it fails since there's no information about the reference beam optics. I agree the solution I applied is not very good, so I welcome further suggestions!

tdimiduk:
Oh, that was silly of me to forget that. Though it looks like calc_intensity doesn't need optics anywhere except in calc_field? 

You can skip the extra call to interpret_args by calling theory._calc_field directly, (all calc_field does is call interpret_args then call theory._calc_field)

vnmanoharan:
@tdimudik and @barkls - it's not clear to me that we want to do this.  The polarization breaks the symmetry of the sphere.  At small angles, comparable to what we detect in a typical holography experiment, there should be no effect of the polarization. However, at large angles, there will be asymmetries in the hologram.  Also, for scattering calculations, which are typically performed over a wide range of angles, there will be differences depending on the polarization. So the user really has to specify a polarization, or the physics is incorrect.

tdimiduk:
@vnmanoharan, right, that would be why I was nervous about that. 

I am considering a shift to storing the polarization with the wavelength in an illumination object, or having a separate polarization argument to the functions, so that we can have what is now the optics object be specialized to things about the optical train (and be optional for all the basic computations)

vnmanoharan:
I don't have a strong opinion on where the polarization argument goes, but if we want to make it an optional input, the way to interpret a function call in which the polarization is not specified is to assume the user wants the calculation done for unpolarized light. In that case, the calculation would average the results for two orthogonal polarizations.

barkls:
Thanks for clearing this up. I now think we should keep polarization as a mandatory argument.

Issue #63, opened 2016-09-07
Multisphere should be able to handle Sphere scatterer


barkls:
If Multisphere is passed a sphere scatterer, it should be able to wrap it in a Spheres() object.

tdimiduk:
Fixed in PR #73

PR #73, opened 2016-09-15
Allowed Multisphere theory to accept Sphere scatterers

barkls:


barkls:
This addresses issue #63

tdimiduk:
You can use the notation:
fixes #63 
to cause github to close the issue when you merge the pr (not sure if it will do it for merging to a feature branch rather than master, but worth getting in the habit of using it.

PR #74, opened 2016-09-16
Move to python 3

tdimiduk:
This PR converts all of holopy to python 3.

I had to switch to nose2 for the unit tests, for some reason nosetests was not running correctly, at least on my machine.

The doctests may not work, @barkls or I may have to fix them. I have not converted any of the documentation to python3, though I don't think there will be many changes.


barkls:
Looks like Travis is failing to import tifffile with conda. Can you try pip installing it?

vnmanoharan:
Do we need tifffile anymore?  The reason we included it originally was to handle 12-bit TIFF files, which PIL couldn't handle.  But my understanding is that pillow can handle them, so no need for tifffile.

PR #75, opened 2016-09-19
Added interpret_args & other minor fixes

barkls:


vnmanoharan:
This PR seems to include a patch that allows the user to call a theory for a sphere without specifying a polarization.  This behavior could lead to problems.  If the user doesn't specify a polarization for a sphere object, the code should either (a) throw an exception or (b) assume unpolarized light.  But it should not assume an arbitrary value for the polarization.

barkls:
Failure to specify a polarization for scattering calculation will correctly throw an exception, just like failure to specify wavelength or medium index. 

I thought I had reverted the earlier commit that arbitrarily assumed polarization, but it seems like I inadvertently continued working on that branch. That specific chunk of code is gone in any case.

holopy/core/io/io.py
```diff
@@ -64,35 +64,25 @@ def load(inf, spacing = None, optics = None, channel=None):
...
-            optics = Optics(**optics)
 
     loaded_yaml = False
+    # attempt to load a holopy yaml file
     try:
-         loaded = serialize.load(inf)
-         loaded_yaml = True
+        loaded = serialize.load(inf)
+        if spacing is not None or wavelen is not None or index is not None or polarization is not None or optics is not None:
+            warn("WARNING: If you are trying to overwrite hologram parameters, you must do so explicitly. Extra arguments are being ignored.")
```
tdimiduk: I don't understand your logic here. What do you mean by overwrite explicitly? I would have thought providing specific arguments with values woudl count as an explicit instruction

holopy/core/io/io.py
```diff
@@ -64,35 +64,25 @@ def load(inf, spacing = None, optics = None, channel=None):
...
 
     loaded_yaml = False
+    # attempt to load a holopy yaml file
     try:
-         loaded = serialize.load(inf)
-         loaded_yaml = True
+        loaded = serialize.load(inf)
+        if spacing is not None or wavelen is not None or index is not None or polarization is not None or optics is not None:
+            warn("WARNING: If you are trying to overwrite hologram parameters, you must do so explicitly. Extra arguments are being ignored.") 
+        loaded_yaml = True
```
tdimiduk: I don't understand the logic here. It looks like you are saying that you are ignoring the arguments the user supplies, but in the code that follows it looks like you do you them.

Also, what do you mean by "explicitly"? Something other than passing them directly as arguments to the function?

holopy/core/metadata.py
```diff
@@ -98,6 +101,14 @@ def wavelen_in(self, medium_index):
         return self.wavelen/medium_index
 
     @property
+    def polarization(self):
+        return self._polarization
+
+    @polarization.setter
+    def polarization(self, val):
+        self._polarization = np.array(val)
+
```
tdimiduk: What is the logic behind using a property/setter here rather than just a simple variable?

You have different behavior with None in the constructor and setter, you should probably move the None check into the setter and then use the setter in the constructor (unless setters don't work in constructors, I can't remember for sure)

holopy/core/metadata.py
```diff
@@ -233,3 +244,21 @@ def theta_phi(self, origin=None):
             for j, phi in enumerate(self.phi):
                 pos[i*self.shape[1]+j] = theta, phi
         return pos
+
+def interpret_args(schema=None, index=None, wavelen=None, polarization=None, optics=None):    
+    from .marray import Schema
+    if not isinstance(schema, Schema):
+        schema = Schema(positions=schema)
```
tdimiduk: Will this behave correctly if you are loading an image? The schema positions argument is meant to be a list of coordinates, here I think you can end up setting it equal to a list of pixel values.

holopy/core/metadata.py
```diff
@@ -233,3 +244,21 @@ def theta_phi(self, origin=None):
...
+        schema.optics=Optics()
+
+    if index is not None:
+        schema.optics.index = index
+    if wavelen is not None:
+        schema.optics.wavelen = wavelen
+    if polarization is not None:
+        schema.optics.polarization = polarization
+    return schema
+
```
tdimiduk: I usually prefer making a new object and setting arguments in the constructor as opposed to the property setting you are doing here. Python passes things by reference, so if you have an existing optics on your schema, this code will change it, which could suprise you if you use the same schema in different places and only set the keywords in some of them.

holopy/core/io/io.py
```diff
@@ -30,11 +30,11 @@
...
 from holopy.core.marray import Image, arr_like
-from holopy.core.metadata import Optics
+from holopy.core.metadata import Optics, interpret_args
 from holopy.core.helpers import _ensure_array
 
 
-def load(inf, spacing = None, optics = None, channel=None):
+def load(inf, channel=None, spacing=None):
     """
     Load data or results
```
tdimiduk: I think you may want to delete the spacing and channel arguments here as well, I think basically the same arguments we made with respect to optics apply

PR #79, opened 2016-09-23
updated docs framework for python 3

barkls:
Also wrote a thorough tutorial on loading images (with built-in doctest) that we can hopefully use as a model for the rest of the documentation.


holopy/core/holopy_object.py
```diff
@@ -61,10 +62,14 @@ class HoloPyObject(Serializable):
 
     @property
     def _dict(self):
-        return dict(self._iteritems())
+        dictlist = dict(self._iteritems())
+        if self.__class__.__name__ is not 'Optics':
+            dictlist = dict_without(dictlist, ('wavelen', 'index', 'polarization'))
+        return dictlist
```
tdimiduk: I don't think we want to be doing something this invasive way up at the root HoloPyObject. If you need to do something like this, override the _dict method at the point where this becomes necessary (probably this is for marray or schema?)

holopy/core/io/io.py
```diff
@@ -69,19 +70,20 @@ def load(inf):
...
-    if not loaded_yaml:
-        pass
-        #TODO load a tif with metadata
-        #TODO confirm tif has metadata
-        #TODO if not, raise exception referring user to load_image function.
-
-    return loaded
+        if os.path.splitext(inf)[1] in tiflist:
+            im = load_image(inf)
+            metadat = yaml.load(pilimage.open(inf).tag[270][0])
```
tdimiduk: Is there a reason you are calling this metadat? It looks kind of like you left off a character. I would probably either spell out metadata, or if you want to shorten, just call it meta here.

holopy/core/io/io.py
```diff
@@ -151,13 +153,77 @@ def save(outf, obj):
...
+            depth = depth-1
+            typestr = 'int' + str(depth)
+        else:
+            raise Error("Unknown image depth")
+            
+        if im.max() <= 1:
+            im = im * ((2**depth)-1) + .499999
+            im = im.astype(typestr)
+    if os.path.splitext(filename)[1] in tiflist:
+        metadat = yaml.dump(im._dict)
```
tdimiduk: I think we may want to look into using json instead of yaml for metadata like this. Don't need to fix that before merging this code though.

holopy/scattering/__init__.py
```diff
@@ -39,5 +39,9 @@
...
 '''
 
-from . import theory
-from . import scatterer
+from . import scatterer, theory
+from .theory import Mie, Multisphere, dda
+from .scatterer import Sphere, Spheres, JanusSphere, Ellipsoid, Capsule, Cylinder, Bisphere
+from .scatterer import Scatterer, Scatterers, Indicators, sphere_builtin, Union, Difference, Intersection
+from .calculations import calc_holo, calc_field, calc_intensity, calc_cross_sections, calc_scat_matrix
+
```
tdimiduk: Thanks for finally extending this import list. I have been meaning to do that for a while and just somehow never got around to it

Is there a reason you are importing lowercase dda (the file) rather than DDA (the theory object)?

I would like to rework some of the dda stuff which may affect these imports, but we can always change it here later.

PR #81, opened 2016-10-07
Create and organize holopy core.tools

barkls:


barkls:
I don't know why the list of commits is so huge and includes changes that already merged. Is there something I should be doing differently?

tdimiduk:
Not sure why the list of commits is so huge either, maybe you started this from a not fully merged branch? In any event, as long as the code changes are correct, I don't think it matters too much. 

I don't think there is a lot of purpose in using the helpers namespace, I would probably expose everything at the core.tools level (ie import them all into __init__.py. I think that file might also be better named something like utilities at this point, but I am less sure about that.

Is there a reason for the separate file for Fourier stuff? I would think those functions could be in math or img_proc, though I guess there isn't really any problem with them being in a separate file.

PR #83, opened 2016-10-11
reduce number of exception classes

### LINKED ISSUE #80: Remove old nose dependency
/holopy/scattering/theory/dda.py", line 40, in <module>

    from nose.plugins.skip import SkipTest

that is using old nose, need to find an equivalent way of doing things with new nose

### LINKED ISSUE #82: Too many different exceptions
We should think about coarse-graining our exceptions.  In general having exceptions for particular errors is useful only insofar as the user might want to catch that particular exception.  But we don't want a different exception for anything that can go wrong, since we have to write all those exceptions and remember their names to import them.  There are several cases in which we can coarse-grain to make it easier to write new functions and classes without significantly impacting the user.  For example, the Multisphere code has three different exceptions:

```
class MultisphereFieldNaN(UnrealizableScatterer):
    def __str__(self):
        return "Fields computed with Multisphere are NaN, this probably "
        "represents a failure of the code to converge, check your scatterer."


class MultisphereExpansionNaN(Exception):
    def __str__(self):
        return ("Internal expansion for Multisphere coefficients contains "
                "NaN.  This probably means your scatterer is unphysical.")

class ConvergenceFailureMultisphere(Exception):
    def __str__(self):
        return ("Multisphere calculations failed to converge, this probably means "
"your scatterer is unphysical, or possibly just huge")
```

All of which tell the user the same thing -- that there is a problem with the scatterer. Is there a way to reduce these to one exception, and use the message to indicate where the point of failure was?  Also, in general, can we reuse exception classes from the standard library (like ValueError) where we are doing standard tasks like validating input?  All of this would make the code easier to read and maintain.

barkls:
fixes #82


barkls:
most recent commits re-introduced dependency on nose to non-test code.

PR #85, opened 2016-10-18
moved an OpenOpt test from test_minimzer to its dedicated file

barkls:


vnmanoharan:
Should we get rid of the OpenOpt code?  I don't think it is widely used, and if we're going to work on the fitting infrastructure, I'd rather it be based on [lmfit](https://github.com/lmfit/lmfit-py/).

barkls:
This was a trivial change, but I've opened a new issue #86 for us to discuss fitting in holopy more generally.

PR #87, opened 2016-10-18
make test_iter_limit actually compare against a gold

barkls:


barkls:
previously it was just checking that the correct warning was raised. I'm merging this directly once travis is finished since this is what was tripping up conda-forge.

Issue #78, opened 2016-09-22
Rethink saving to file


barkls:
hp.save()
-defaults to saving any holopyobject as hdf5
-option to save to yaml if preferred

hp.save_image()
-saves Image object to image file
-saves metadata if compatible format
-options to control contrast range, discretization

barkls:
current status: 
hp.save_image() works
hp.save() only does yaml and tif

tdimiduk:
hp.save now saves anything xarray derived (so anything with potentially large amounts of array data) as hdf5.

Issue #45, opened 2016-08-19
switch to xarray


barkls:
Pandas is currently used in the Bayesian inference code. We might as well expand it to the rest of holopy where it makes sense, for example timeseries data.

vnmanoharan:
We should look into using [xarray](http://xarray.pydata.org/en/stable/why-xarray.html). Unlike pandas, it supports multidimensional data sets.  It also [supports metadata](http://xarray.pydata.org/en/stable/data-structures.html#dataarray) in that you can assign a dictionary containing the metadata to a data set. Finally, it is based on pandas, so that we would not have to rewrite the Bayesian inference code.  We might think about using it to replace Marray.

tdimiduk:
Yes, it looks like xarray is going to be the right way to do things. I have a very solid proof of concept that it works in my xarray branch:
https://github.com/tdimiduk/holopy/tree/xarray

I looks like it will be able to replace Marray, and give a much cleaner way of handling things like electric field components and timeseries or z stacks. 

I should be submitting a PR soon.

tdimiduk:
Backwards comparability question I would like @vnmanoharan's opinion on. 

xarray looks like it will make it possible to eliminate most of the marray classes and code. However, loading old yaml files needs those classes to exist. So far I see two decent ways of dealing with this:

1) Eliminate the code from holopy, provide a separate tool to convert old files into new hdf5/netcdf files

2) Keep the code in holopy but separate it off into a backwards compatibility file/directory so that holopy can still read all the old files directly, constructing new xarray objects for the user (which in turn can/should be saved as new hdf5/netcdf)

Opinions on which route to take, or do you want to propose a third option?

vnmanoharan:
Spoke to @tdimiduk offline about this, but re-capping for the record: my preference is to get rid of the legacy code, since otherwise we'll have to maintain it, and to provide a separate tool to convert between formats.

PR #91, opened 2016-10-19
Use xarray instead of marray

tdimiduk:
This PR still has some issues and lots of failing tests, but it is a substantial change to the core of how we do things, so I wanted to put it up for comment. It may also be worth merging (starting an xarray branch in the main holopy repo) so that @barkls can work on things as they are going to be. 

As suggested in #45 this PR implements a switch of HoloPy's core array handling to use the xarray library. This allows us to get rid of Marray and most of our position handling code.

Xarray also gives us straightforward serialization to/from netcdf/hdf5, which will provide the core of the remaining work for #78.


holopy/propagation/convolution_propagation.py
```diff
@@ -60,10 +57,14 @@ def propagate(data, d, index=None, wavelen=None, gradient_filter=False):
...
-    data = interpret_args(data, index, wavelen)
-
     if data.positions is None:
         raise MissingParameter("image spacing")
+    if wavelen is None:
+        wavelen = data.illum_wavelen
+    if index is None:
+        index = data.med_index
+    med_mavelen = wavelen/index
+
```
barkls: This looks like another opportunity to use optical_parameters. We can then use get_med_wavelen. Additionally, the MissingParameter check (below) should be updated for med_index and illum_wavelen.

holopy/core/metadata.py
```diff
@@ -142,123 +45,112 @@ def __init__(self):
     def __str__(self):
         return ("Medium index not specified in Optics instance.")
```
barkls: This is mostly unrelated to your PR, but these exception classes shouldn't be living in metadata. Actually, it looks like they aren't used anywhere in holopy so you should delete them (they are redundant with MissingParameter in any case)

holopy/scattering/calculations.py
```diff
@@ -23,20 +23,21 @@
...
+from holopy.core.metadata import vector, Image, optical_parameters
 from holopy.core.tools import dict_without, is_none
 from holopy.scattering import Mie, Multisphere, Sphere, Spheres
 from holopy.scattering.theory import dda
 from holopy.scattering.errors import AutoTheoryFailed, MissingParameter
 
 import numpy as np
+from xarray.ufuncs import square
 
 def check_schema(schema):
```
barkls: All uses of check_schema have been removed in this PR, although I'm not sure the functionality has been maintained. Goal is that someone can call scattering functions with optical parameters specified in either schema or explicitly as scattering function arguments, but they get a MissingParameter if not specified in either place.

In any case, check_schema should be either used in the scattering functions or deleted.

holopy/scattering/theory/multisphere.py
```diff
@@ -39,6 +40,8 @@
                       MultisphereFailure)
 from .scatteringtheory import FortranTheory
 
+def normalize_polarization(illum_polarization):
```
barkls: Is there any chance of this being used somewhere else? It seems like it could be useful to have it more accessible than inside multisphere...

holopy/core/metadata.py
```diff
@@ -142,123 +45,112 @@ def __init__(self):
...
+
+def angles_list(theta, phi, medium_index, illum_wavelen, illum_polarization, normals=(0, 0, 1)):
+    # This is a hack that gets the data into a format that we can use
+    # elsewhere, but feels like an abuse of xarray, it would be nice to replace this with something more ideomatic
+    d = make_attrs(medium_index, illum_wavelen, illum_polarization, normals)
+    d['theta'] = theta
+    d['phi'] = phi
+    return xr.DataArray(np.zeros(len(theta)), dims=['point'], attrs=d)
+
+def ImageSchema(shape, spacing, medium_index=None, illum_wavelen=None, illum_polarization=None, normals=(0, 0, 1)):
```
barkls: Obviously that won't work because we need information about image size. I still think it would be nice to get rid of ImageSchema, but I can't think of a smart way to do it right now.

tdimiduk: I agree we probably want to get rid of ImageSchema. ImageSchema as it stands is basically a transitional porting tool (so I didn't have to go change every place where ImageSchema is used). Once we figure out what we want the final interface to be, we can replace it in the code and the docs.

holopy/scattering/calculations.py
tdimiduk: The way I have structured the new code I believe what will happen if you are missing one of the parameters is you will get an error about a missing parameter to one of the internal scattering functions. Which is probably not ideal as a user experience. I think what might make sense is to do some kind of checking in the optical_parameters function since that is used in all the right places and is the place that unifies data from the schema and function arguments.

Currently it would be a little difficult because I purposefully structured the optical_parameters function to gracefully handle whatever data you ask it for. However, I think there are only a small set off different cases (mostly whether you need polarization or not) so we could have two or three similar functions that do the checking and raise helpful errors.

holopy/core/io/io.py
```diff
@@ -258,8 +247,7 @@ def load_average(filepath, refimg=None, wavelen=None, index=None, polarization=N
     if len(filepath) < 1:
         raise LoadError(filepath, "No images found")
 
-    refimg = interpret_args(refimg, spacing, wavelen, index, polarization, optics)
-    accumulator = load_image(filepath[0], refimg.spacing, optics=refimg.optics)
+    accumulator = load_image(filepath[0], refimg.spacing, wavelen, index, polarization)
```
tdimiduk: Optical parameters as it is currently specified is not the right tool for this because it does conversions to wavevec. You may be right that a simplified version of interpret_args may need to come back, I hacked the old one out because it was mostly doing marshalling of optics objects, but some of the functionally might still be needed here.

holopy/core/io/io.py
```diff
@@ -40,31 +40,34 @@
 
 tiflist = ['.tif', '.TIF', '.tiff', '.TIFF']
 
+def default_extension(inf, defext='.h5'):
+    file, ext = os.path.splitext(inf)
+    print(file, ext)
+    print(type(ext))
```
barkls: Are these prints intentional?

barkls: Are these prints intentional?

tdimiduk:
No those are debugging prints I forget to delete

PR #94, opened 2016-10-20
V3

barkls:


tdimiduk:
I am not sure we want to add the full_data files to the setup script. They are mostly for us in development to increase confidence, but I don't think we need to ship them with builds.

Also what are the changes to the binary files recon_*.npy?

barkls:
sorry for the confusion. This is a duplicate of PR#92 with one additional commit to fix a test that didn't get merged correctly.

barkls:
recon_*.npy aren't used by anything anymore.

barkls:
I agree we don't need to ship full_data files in setup script. I will revert that in my next PR.

PR #100, opened 2016-10-25
fix broken propagation test

barkls:
I used a.dims.index('x') to find the dimension of a that is labelled 'x'. If there's a better way to do it let me know.


barkls:
ooh didn't mean to include that setuptools commit. i'll remove it

barkls:
well it's not showing up in the code difference, just in the commit notes, so I guess it's ok

tdimiduk:
Good find on this bug

PR #109, opened 2016-11-03
Io & docs

barkls:
This PR includes lots of changes to sphinx-generated apidoc .rst files. These are still in flux, so don't bother going through them. There are real changes, esp. to metadata.py and io.py


docs/source/users/install.rst
```diff
@@ -84,10 +84,87 @@ to get you started.
...
+
+#. Install gfortran. Follow the instructions to install the
+   appropriate binaries from HPC `here
+   <http://hpc.sourceforge.net/>`_. (these aren't available through
+   macports)
+
+#. Install the required python packages one of two ways
+
+   #. **With prepared packages** 
+
```
tdimiduk: Will eventually want to switch this to anaconda probably

holopy/core/io/io.py
```diff
@@ -283,8 +293,29 @@ def load_average(filepath, refimg=None, wavelen=None, index=None, polarization=N
...
+    if not (raw.shape == bg.shape == df.shape and list(get_spacing(raw)) == list(get_spacing(bg)) == list(get_spacing(df))):
+        raise BadImage("raw and background images must have the same shape and spacing")
 
-    return accumulator/len(filepath)
+    raw = deepcopy(raw)
+    bg = deepcopy(bg)
+    raw -= df
+    bg -= df
+    raw /= bg
+    return raw
```
tdimiduk: Consider switching to using regular arithmatic internally and then returning copy_metadata(raw, result)

holopy/core/metadata.py
```diff
@@ -88,15 +87,22 @@ def make_coords(shape, spacing, z=0):
...
     return {'x': np.arange(shape[0])*spacing[0], 'y': np.arange(shape[1])*spacing[1], 'z': 0}
 
-def make_attrs(medium_index, illum_wavelen, illum_polarization, normals=None):
-    if is_none(normals):
-        normals = (0, 0, 1)
-    return {'medium_index': medium_index, 'illum_wavelen': illum_wavelen, 'illum_polarization': to_vector(illum_polarization), 'normals': to_vector(normals)}
+def update_metadata(a, medium_index=None, illum_wavelen=None, illum_polarization=None, normals=None):
+    attrlist = {'medium_index': medium_index, 'illum_wavelen': illum_wavelen, 'illum_polarization': to_vector(illum_polarization), 'normals': to_vector(normals)}
+    a.attrs = updated(a.attrs, attrlist)
+
```
tdimiduk: consider adding an overwrite option (probably defaulting to False), and make a shallow copy before modifying then return the copy.

holopy/core/io/io.py
```diff
@@ -283,8 +293,29 @@ def load_average(filepath, refimg=None, wavelen=None, index=None, polarization=N
...
+
+    if is_none(df):
+        df = ImageSchema(raw.shape,get_spacing(raw))
+
+    if not (raw.shape == bg.shape == df.shape and list(get_spacing(raw)) == list(get_spacing(bg)) == list(get_spacing(df))):
+        raise BadImage("raw and background images must have the same shape and spacing")
 
-    return accumulator/len(filepath)
+    raw = deepcopy(raw)
+    bg = deepcopy(bg)
```
tdimiduk: Deepcopy can be kind of expensive (both computation and memory) I would try to avoid it (see next comment)

holopy/core/io/io.py
```diff
@@ -131,15 +130,18 @@ def load_image(inf, spacing=None, illum_wavelen=None, medium_index=None, illum_p
         else:
             arr = arr[:, :, channel]
     elif channel is not None and channel > 0:
-        warnings.warn("Warning: not a color image (channel number ignored)")
+        warn("Warning: not a color image (channel number ignored)")
```
tdimiduk: Does load image do something smart with color channels and xarray? I don't think so yet, but it would be a good thing to add.

PR #111, opened 2016-11-10
reorganize tools and metadata. 

barkls:
Updated file io so that objects maintain metadata in the same format.
Changed how scattering calculations pass optical arguments around.


holopy/scattering/calculations.py
```diff
@@ -124,10 +120,10 @@ def calc_holo(schema, scatterer, medium_index=None, illum_wavelen=None, illum_po
         Calculated hologram from the given distribution of spheres
     """
     theory = interpret_theory(scatterer,theory)
-    par = optical_parameters(schema, medium_index=medium_index, illum_wavelen=illum_wavelen, illum_polarization=illum_polarization)
-    scat = theory._calc_field(scatterer, schema, **par)
-    holo = scattered_field_to_hologram(scat*scaling, par['illum_polarization'], schema.normals)
-    return finalize(schema, holo)
+    uschema = update_metadata(schema, medium_index, illum_wavelen, illum_polarization)
```
tdimiduk: Why are you using update_metadata in some places and optical_parameters in others?

holopy/scattering/theory/scatteringtheory.py
```diff
@@ -42,7 +45,8 @@ class ScatteringTheory(HoloPyObject):
     implementing a _calc_field(self, scatterer, schema) function that returns a
     VectorGrid electric field.
     """
-    def _calc_field(self, scatterer, schema, medium_wavevec, medium_index, illum_polarization):
+    
+    def _calc_field(self, scatterer, schema):
```
tdimiduk: We should talk about this, I thought we had reached a different conclusion about how we are passing in the metadata than it looks like you have done here. I am not sure you are not right, but I would like to chat about it before making this change.

Issue #43, opened 2016-08-19
Bayesian Inference


barkls:
Interface needs some thought. Docs need heavy updating to emphasize its existence.

barkls:
Separate Inference directory from fitting directory. Probably have to move some stuff to core that they both depend on.

barkls:
A couple docs comments:

1) code should all run in < 1min. Emphasize that these are not parameters to use for actual data
2) noise_sd variable should be better assigned - use add_noise function
3) doctest.

PR #115, opened 2016-11-17
New Inference Interface

### LINKED ISSUE #50: Store Initial conditions with subset tempering
SamplingResult should have an initial_condition property that returns initial walker positions. This needs to be preserved when saving with a specified burn-in.

tdimiduk:
Standardizes interface between normal and subset tempered samplings. Partially addresses #43 

Adds Strategy objects to describe how an inference is run. These are stored as part of the result and should contain the information needed to do the same sampling again.

Store sampling results in an xarray Dataset with original data. Serialize model and strategy into the same hdf5 file. For subset tempering, also store results from all the tempering stages. Fixes #50


holopy/inference/tests/test_mcmc.py
```diff
@@ -60,8 +52,6 @@ def test_subset_tempering():
     mod = AlphaModel(scat,noise_sd=.1, alpha=prior.Gaussian(0.7,0.1))
     with warnings.catch_warnings():
         warnings.simplefilter('ignore')
-        inf=mcmc.subset_tempering(mod,holo,final_len=10,nwalkers=4,stages=1,stage_len=10,threads=None, verbose=False,seed=40)
-    assert_obj_close(inf.most_probable_values(),gold_alpha, rtol=1e-5)
-    assert_equal(inf.n_steps,gold_nsteps)
-    assert_obj_close(inf.acceptance_fraction,gold_frac)
-    assert_obj_close(float(inf.data_frame(burn_in=6)[1:2].alpha),gold_alpha, rtol=1e-1)
+        #inf=mcmc.subset_tempering(mod,holo,final_len=10,nwalkers=4,stages=1,stage_len=10,threads=None, verbose=False,seed=40)
```
barkls: remove commented out line

tdimiduk:
New commits actually have subset tempering saving correctly now. 

I still need to do some more testing of the sampling framework to confirm that it is reproducing old results, but I think this PR is ready to merge. I'll start a new one with more tests and fixes if I find any correctness issues.

PR #116, opened 2016-11-17
Rename former schema objects

barkls:


barkls:
Also deleted some unused code that has not been maintained leading up to 3.0 release.

PR #117, opened 2016-11-17
add back in storing seperate x, y, and z coords when flattening

tdimiduk:
makes coordinate calculations ~5x faster


barkls:
All set to merge. "Pull requests that have a failing status can’t be merged on a phone."

PR #119, opened 2016-11-26
implement detector_points

barkls:
This replaces detector_far and is more flexible.

Detectors no longer contain optical information when created.

Generalized 'flat' and 'point' dimensions in a few places.


docs/source/pyplots/calc_two_spheres.py
```diff
@@ -1,15 +1,14 @@
 import holopy as hp
 import numpy as np
-from holopy.core import ImageSchema
-from holopy.scattering import calc_holo, Sphere, Spheres
+from holopy.scattering import calc_holo,Sphere, Spheres
```
tdimiduk: missing space after calc_holo

holopy/core/errors.py
```diff
@@ -35,3 +35,7 @@ class BadImage(Exception):
 class NoMetadata(Exception):
     def __str__(self):
         return "File without metadata detected. To load raw images, use hp.load_image()"
+
+class CoordSysError(Exception):
+    def __str__(self):
+        return "Could not interpret your points. Use either Cartesian or spherical coordinates"
```
tdimiduk: Consider if this new exception pulls its weight? Could it be a ValueError? Could you unify it with some other existing exception?

holopy/core/metadata.py
```diff
@@ -134,22 +142,45 @@ def from_flat(a):
...
     return a
 
-def sphere_coords(a, origin=(0,0,0), wavevec=1, include_r=True):
-    if hasattr(a,'theta') and hasattr(a, 'phi') and not include_r:
-        # we are working in far field
-        return {'theta': a.theta, 'phi': a.phi}
+def sphere_coords(a, origin=(0,0,0), wavevec=1):
+    if hasattr(a,'theta') and hasattr(a, 'phi'):
+        out = {'theta': a.theta.values, 'phi': a.phi.values, 'point':a.point.values}
+        if hasattr(a, 'r') and any(a.r < np.inf):
```
tdimiduk: Probably consider using np.isfinite
https://docs.scipy.org/doc/numpy/reference/generated/numpy.isfinite.html
floating point infinities can be tricky beasts sometimes so the numpy function is probably more reliably correct.

holopy/scattering/calculations.py
```diff
@@ -51,7 +56,7 @@ def interpret_theory(scatterer,theory='auto'):
 
 def finalize(schema, result):
     if not hasattr(schema, 'flat'):
-        result = from_flat(result)
+        result=from_flat(result)
```
tdimiduk: Convention is to put spaces around binary operators (like =) except in when assigning to function arguments inside the parens, so probably put these spaces back unless you had a good reason for removing them.

Issue #136, opened 2016-11-29
apidocs not working on readthedocs


barkls:
We have no docstrings! or code...

barkls:
addressed by PR #147.

PR #147, opened 2016-12-09
Docs

barkls:
This PR allows readthedocs to run holopy code even though it can't compile fortran. 

setup.py doesn't try to compile fortran if it is running in a virtualenv.


barkls:
looks like my two PRs will have a merge conflict for the import dda line of calculations.py

barkls:
Fixes #136.

We'll need to add image files for scattering results, since it can't execute the pyplots scripts for scattering.

barkls:
On second thought, this might cause problems with compiling HoloPy on conda-forge...We will have to figure out a way to distinguish the readthdocs virtualenv from conda-forge virtualenv, or somehow find out if the fortran compiling will work beforehand.

barkls:
On third thought, travis did work (except for the display fail addressed in PR #146) so we might be ok on conda-forge, since it's also running on a CI system.

PR #150, opened 2017-01-05
Fix xarray saving bug

barkls:
fixed a bug where attrs with a value of None (eg polarization) caused errors when saving.


barkls:
Visualization test on travis is still expected to fail. all tests succeed locally.

barkls:
Included in PR #151.

PR #154, opened 2017-02-21
V3

barkls:


barkls:
All tests passed locally. It's not obvious to me why multisphere tests are failing on travis, but I'm not going to debug it right now.

barkls:
All nosetests now pass. Travis failed because doctests are not up to date.

PR #183, opened 2017-06-13
Make sure all links work in user guides.

barkls:


vnmanoharan:
A couple of corrections needed:
1) in calc_tutorial: Should read "T-matrix" and not "transfer matrix"  The T-matrix relates the incident field to the scattered field.  Transfer matrices are a different approach.
2) typos in "indicies" and "dumbell" in dda_tutorial.rst

barkls:
just saw these, will address.

Issue #180, opened 2017-06-13
Visualization update


barkls:
Remove 3d Visualization code, or tag it as experimental. 
Mayavi hasn't been tested with python 3.
It's not clear whether or not this code would work with the most recent version of HoloPy.

Ellen has done some work on this.

barkls:
Fixed by #187.

PR #187, opened 2017-06-14
Removed 3D visualization because mayavi isn't supported for python 3.

ellenklein:


barkls:
Fixes #180.

@vnmanoharan there was some discussion of keeping the 3d visualization code and marking it as experimental. There is not guarantee that this code will work with HoloPy 3 or even Python3, so my preference would be to remove it for now and open an issue about 3d visualization that links to the commit where it was removed. What is your opinion?

vnmanoharan:
The API for mayavi and mlab has not changed, to the best of my knowledge, so the code as written should be fine.  My preference is to keep vis3d.py (rather than deleting it) and mark it as experimental, but to make the changes that @ellenklein made to show.py so that the functions in vis3d.py are never called unless the user explicitly calls them.

tdimiduk:
@vnmanoharan That seems like a reasonable course of action.

PR #188, opened 2017-06-14
Add Windows SkipTests

barkls:
Skip a test to run multicore DDA since it asks user for login credentials and that cannot be automated.

Skip a test that calls scipy.special.spherical_jn and scipy.special.spherical_yn with complex numbers since they are not compiled properly on Windows for conda.


barkls:
@ralex0 please confirm these act as intended on a windows system.

ralex0:
Hey Solomon, it works as intended.

PR #189, opened 2017-06-15
Add developer docs for adding an inference model

tdimiduk:
Also rename _holo in NoiseModel to _forward since NoiseModel can work with
things that are not holograms.


tdimiduk:
I wonder if we should rename scatterer in NoiseModel as well? The model is not exactly specific to just scatterers.

PR #190, opened 2017-06-15
Improved docstrings for Mie and Multisphere scattering code. Added se…

jeromefung:
…veral references to credits that were previously in code comments.


barkls:
Looks good Jerome, thanks for the detailed docstrings!

PR #197, opened 2017-07-02
Tmatrix

vnmanoharan:
merge in-progress T-Matrix theory for axisymmetric scatterers


jeromefung:
The level of agreement between Tmatrix and Mie is likely due to the fact that Mie (and Multisphere) incorporate non-asymptotic r dependence of the scattered field, whereas both Tmatrix and DDA as implemented assume a far-field spherical wave r dependence (E ~ e^{ikr}/kr) .

barkls:
Yes, I guessed far-field might have something to do with it. @anna-wang did you ever have the tmatrix test passing at the previous tolerance level?

anna-wang:
Solomon - I don't think I ran that same test previously. Jerome is right
about the agreement (or lack thereof); the agreement should get worse with
a decrease in z.

Anna

barkls:
merging now to facilitate comparisons when adding more changes.

The generalization of _raw_scat_matrs  in v3.1 will conflict with the changes here, so it makes sense to merge all changes to date into one branch.

Issue #181, opened 2017-06-13
Windows conda build


barkls:
Hopefully we can get this working for non-developers wanting to install on Windows.
We should also update the windows installation documentation.

Ron is working on this.

barkls:
@ralexander

barkls:
@ralex0

ralex0:
I was able to successfully build and install a holopy Anaconda package (preferred built distribution format as suggested by @vnmanoharan) on my win-64 python 3.6 environment.

But I suppose I should wait until all the other issue on the milestone have been closed before doing the official build, right?

vnmanoharan:
We should test it before the final release, to make sure it works on other (Windows) machines.  Not sure if GitHub has a mechanism for doing this, so you might want to email it to me and @barkls to test.

jeromefung:
Hi Ron --

Out of curiosity, what ended up being the solution to the issue with the
scipy.special spherical Bessel functions we were running into yesterday?

Best,
Jerome

barkls:
fyi conda-forge uses a continuous integration platform called appveyor to test Windows builds. 

I don't know how easy it is to set up, or if it is even free, so I agree the best option is to send it to me and Vinny. 

Jerome: I'm not certain what the final resolution was, so I defer to @tdimiduk

ralex0:
I'm not allowed to upload the file to gmail because it's executable code, but zipping it and uploading here works.

[holopy-3.0-0.tar.zip](https://github.com/manoharan-lab/holopy/files/1072794/holopy-3.0-0.tar.zip)

I believe all you have to do is drop this in `~\Anaconda3\conda-bld` or equivalent, then do `conda install --use-local holopy`

ralex0:
@jeromefung I think building scipy yourself from the source might fix it, but I have not confirmed. 

If you recall, our issue is a duplicate of scipy/scipy#7002 and it seems they determined that it's a problem with Anaconda's scipy build. The open issue is ContinuumIO/anaconda-issues#1415.

I think for now we decided to skip that test for Windows users. @tdimiduk suggested using `nose.SkipTest`, but I didn't get that working properly yet. This is one of the commits I wanted to make before the final build.

vnmanoharan:
I was able to install it on Windows 10 build 15063 by unzipping first, then installing with "conda install holopy-3.0.0.tar.bz2".  But after running `import holopy` from python, I get the following error:
```
UserWarning:
[...]

PR #199, opened 2017-07-12
windows installation docs

barkls:
It might be better to create a manoharan-lab conda channel.

### DIFF
--- a/docs/source/users/install.rst
+++ b/docs/source/users/install.rst
@@ -24,13 +24,11 @@ If this line works, skip to :ref:`usage` before diving into the tutorials.
 
 Windows Support
 ---------------
-To find a built distribution of HoloPy on Windows, use ::
+To find a channel with a built distribution of HoloPy for Windows, use ::
 
     anaconda search holopy
 
-You will need to reproduce the build environment that compiled the built distribution for libraries to link up properly. 
-Specifically, you should be running the same versions of Windows, python, and numpy.
-You may also need to manually install any missing dependencies (eg. emcee) from conda-forge like ::
+You may need to manually install any missing dependencies (eg. emcee) from conda-forge like ::
 
     conda install -c conda-forge emcee
 
--- a/docs/source/users/install.rst
+++ b/docs/source/users/install.rst
@@ -36,8 +36,17 @@ Optional dependencies for certain calculations include:
 
 Windows Support
 ---------------
-We are working on distributing a Windows HoloPy build that does not require Fortran compilers.
-In the mean-time your best option is to build it yourself by following the instructions for :ref:`dev_install`.
+To install a built distribution of HoloPy on Windows, you will need to reproduce our build environment.
+Specifically, you will need to be running a 64-bit version of Windows 10 with python 3.6 and numpy 1.13 installed.
+You can then try::
+
+conda install -c ralex0 holopy
+
+You may have to install missing dependencies (eg. emcee) from conda-forge like::
+
+conda install -c conda-forge emcee
+
+You can also try to build HoloPy from source by following the instructions for :ref:`dev_install`.
 
 ..  _usage:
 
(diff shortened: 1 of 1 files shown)


vnmanoharan:
Since the channel might change (and we might eventually get conda-forge working), perhaps its best for now to just say that the easiest way to install on windows is to search for a package using `anaconda search holopy` and install that.  I think the docs are the place to put specific info about how to build holopy and general info about how to install binaries (since our methods for producing binaries might change rapidly). The most appropriate place for describing how to install binaries might be in the README.rst

barkls:
I agree it would be nice to not have to update docs whenever we change distribution, so maybe that's not the best place to put it. I'm not sure the readme is a good choice either though, since we want to include specific technical instructions for installing binaries on windows. My interpretation of the readme is that it should be a general entry point to HoloPy for all users. We do have an older file called ``INSTALL.txt`` that hasn't been updated in a while (but is still accurate). That seems like a logical place to put detailed installation instructions.

vnmanoharan:
I don't think INSTALL.txt is the place to put this info, because it's not easily accessible by the web.  There are three webpages where people can get introduced to HoloPy:

1) http://holopy.readthedocs.io/en/master/
2) https://github.com/manoharan-lab/holopy
3) http://manoharan.seas.harvard.edu/holopy

I think (1) should just have a statement about using `anaconda search holopy` and then installing from whatever channel that reports.  Right now, it reports that ralex0 has the win64 binaries and conda-forge for linux and macos.  But I don't htink we should add the channel name to the documentation, since that will likely change.

(2) points to the README.rst, which is not a bad place to put the instructions for installation, or at least a link to the relevant page in the manual.

(3) might be used to give more specific information about which channel to use.  But I think it is sufficient to update (1) with the instructions (generic for all platforms) to do `anaconda search holopy`.

barkls:
I agree that INSTALL.txt is not very accessible. I don't really see the purpose of this file anymore. It does not contain any information that cannot be found in the more accessible documentation.

I updated the docs to tell users to just do ``anaconda search holopy`` and go from there. The instructions are general and won't need to change with future updates to HoloPy or distribution. However, I disagree with this approach because we spent a long time trying to match versions of python and numpy. I think we should provide potential users with the specific information they need to reproduce the build environment. Maybe there's a way to specify this information when uploading to a conda channel?

vnmanoharan:
But this is what dependencies are for. The conda packages need to specify what minimum version of python, numpy, and xarray they will work with.

barkls:
I think there are two different sets of dependencies here. 

HoloPy has a set of dependencies, eg. python 3 & np>=1.12.

My understanding is that a built distribution has a more specific set of requirements, eg. python 3.6 and np=1.13 to reproduce the environment where it was compiled. I thought this was problem in #181. I don't know of a way to get these specific requirements from the conda package unless it has a different set of dependencies than HoloPy itself.

vnmanoharan:
I mean the dependencies in the holopy conda package.  If I try to install a package from a conda channel, and (say) that package requires numpy 1.13 or higher, conda tries to figure out a solution to satisfy all the dependencies of the package I'm installing and the ones I already have. So conda is built to solve exactly this problem. We just need to make sure that the packages have the correct dependencies.

vnmanoharan:
Or, if the conda package inherits these dependencies from HoloPy, we should just update the HoloPy package.  That means we would require python 3.6 and numpy >= 1.13.

barkls:
Well no, that is only a requirement for a specific Windows built
distribution. I think the solution is for Ron to specify additional
dependencies when building a conda package on Windows to exactly replicate
his build environment. Ideally we would have Windows builds available for
multiple versions of python and numpy. For example, the requirement
wouldn't be np>=1.13. It would be np=1.13 for a particular Windows build
only.

vnmanoharan:
Whatever the `np=` criterion, this is a problem that conda is meant to solve, and we should make the packages so that conda can solve it.

vnmanoharan:
The term "built distribution" is confusing to me.  I would call it a "conda package" unless you mean something different.

PR #200, opened 2017-07-18
Tmatrix improvements

barkls:
File comparison is showing some earlier changes that got merged in out of order. Sorry.

Here are the action items I can think of that still need to be addressed to finish the Tmatrix addition:

- Write some documentation (using example_tmatrix as a base)
- Write more extensive tests
- Figure out how to compile Fortran to binary file within setup.py
- Possible change the name, since Tmatrix is ambiguous with Multisphere


vnmanoharan:
Multisphere does not actually do a T-matrix calculation in the way that we use it, so I think the name is fine.

PR #203, opened 2017-07-26
Install tmatrix

barkls:
I don't expect this makefile to work as-is on Windows. For one, ``rm`` should be replaced with ``del``.


barkls:
Note the failing test is fixed already in #202

barkls:
I hopefully fixed the ``rm``/``del`` issue. @ralex0 it would be helpful if you can try this out. Maybe we need to rename ``make`` to ``gmake`` or something in ``setup.py``.

holopy/scattering/theory/tmatrix.py
```diff
@@ -40,11 +40,9 @@ class Tmatrix(ScatteringTheory):
     Computes scattering using the axisymmetric T-matrix solution by Mishchenko
     with extended precision.
 
-    It can calculate scattering from axisymmetric scatterers. Calculations for
-    particles which are very large and have high aspect ratios may not
-    converge.
+    It can calculate scattering from axisymmetric scatterers i.e. cylinders and spheroids. Calculations for
+    particles which are very large and have high aspect ratios may not converge.
```
vnmanoharan: change to read "It can calculate scattering from axisymmetric scatterers such as cylinders and spheroids. Calculations for particles that are very large or have high aspect ratios may not converge."

holopy/scattering/theory/tmatrix_f/makefile
```diff
@@ -0,0 +1,23 @@
...
+	#Windows
+	RM=del S.lp.o lpq.o
+else
+	#Not Windows
+	RM = rm S.lp.o lpq.o
+endif
+
+#make files
+S.exe : S.lp.o lpq.o
+	gfortran -o S.exe S.lp.o lpq.o
```
vnmanoharan: leave off the ".exe" here.  `gfortran -o S S.lp.f lpq.f` will compile the executable and automatically give it the appropriate platform-specific extension (no extension on Linux, .exe extension on Windows).

vnmanoharan:
see comments

barkls:
I think we now have all changes for incorporating Tmatrix scattering theory into HoloPy. Thanks to @ralex0 and @anna-wang for the help!

If there are no more changes after this, we can merge in 3.1 and do another release on conda-forge

Issue #209, opened 2017-10-13
Bayesian inference of Parameter Values for Multiple Spheres


rbnvrw:
First of all, thank you for making this code publicly accessible and thank you for all the work that has gone into this.

I was following the tutorial on "Bayesian inference of Parameter Values" and I am trying to adapt it to multiple spheres. MWE:

```
import holopy as hp
import numpy as np
from holopy.core.io import get_example_data_path, load_average
from holopy.core.process import bg_correct, subimage, normalize
from holopy.scattering import Sphere, calc_holo
from holopy.inference import prior, AlphaModel, tempered_sample
from holopy.scattering import Spheres

s1 = Sphere(center=(5, 3, 5), n = 1.59, r = 1)
s2 = Sphere(center=(5, 5, 5), n = 1.59, r = 1)
s3 = Sphere(center=(5, 7, 5), n = 1.59, r = 1)
collection = Spheres([s1, s2, s3])

medium_index = 1.33
illum_wavelen = 0.405
illum_polarization = (1,0)
detector = hp.detector_grid(shape = 200, spacing = .05)
holo_405 = calc_holo(detector, collection, medium_index, illum_wavelen, illum_polarization)

# Set up the prior
pg5 = prior.Gaussian(4.97, .08)
pn = prior.Gaussian(1.55, .1)
pr = prior.BoundedGaussian(0.98, .05, 0, 5)
s1 = Sphere(center=(pg5, prior.Gaussian(2.97, .08), pg5), n = pn, r = pr)
s2 = Sphere(center=(pg5, pg5, pg5), n = pn, r = pr)
s3 = Sphere(center=(pg5, prior.Gaussian(6.97, .08), pg5), n = pn, r = pr)
priors = Spheres([s1, s2, s3])

noise_sd = holo_405.std()
model = AlphaModel(priors, noise_sd=noise_sd, alpha=1)
result = tempered_sample(model, holo_405, nwalkers=10, samples=100, max_pixels=200*200*0.0005)

result.values()
```

If I run this code, I get a lot of `OverlapWarning`s and the following output:

```
{'0:Sphere.center[1]': UncertainValue(value=2.991103114359165, plus=0.27333182893886043, minus=0.14208225419107867, n_sigma=1),
 '2:Sphere.center[1]': UncertainValue(value=6.8950442696398495, plus=0.1741085966110747, minus=-0.007508585247170352, n_sigma=1),
[...]

Issue #194, opened 2017-06-30
Multisphere calculations return NaN


barkls:
There is a bug that has happened on Travis under a few circumstances (but never observed on an actual machine). 

Some, multisphere calculations return NaN, resulting in test failures. Other multisphere calculations run normally. This bug happened for a while when using numpy 1.11 after upgrading to xarray 0.9. It has also inconsistently occurred on python3.6 with numpy 1.12. 

Strangely, compiling and running identical code can sometimes fail and other times not, which is characteristic of weird memory issues with the F77 multisphere code.

This is a tricky issue to debug since it only occurs on Travis (so far), is not reproducible, and involves old F77 code. Hopefully it is fixed when we eventually upgrade to the F90 MSTM code.

jeromefung:
I agree that this is going to be a pain to debug if it only happens
intermittently and in Travis. Out of curiosity, is it the same test or
tests that fail consistently, or does it seem random? About what fraction
of the time would you estimate this to be happening?

Jerome

barkls:
It is usually the same tests, but not always the same number that fail.
Interestingly, low level tests have always passed, so the code is compiled,
it's just not passing values around properly. On builds that aren't
working, tests fail about 50-80% of the time, so it's not a rare one-off
problem that would go undetected.

barkls:
Note that this is the only reason we had to require np>=1.12. If this is fixed, we can relax that dependency requirement.

barkls:
Recently this has been happening on PR Travis builds but not master Travis builds.

barkls:
@ellenklein encountered this (or a similar issue) on a mac running a conda-forge installation of holopy 3.2.1. She had scipy v1.1.0 and tried a few different numpy versions. She wasn't able to run calc_holo on any Spheres objects.

briandleahy:
This has not appeared for >2  years, and with our current more-extensive test coverage, it seems to have been solved. It may have been a numpy issue.

PR #210, opened 2017-10-13
Add documentation on tied priors to the Bayesian inference tutorial

rbnvrw:
As suggested by @tdimiduk (https://github.com/manoharan-lab/holopy/issues/209#issuecomment-336445072)

Please feel free to edit the text if needed, as English is not my first language.

### DIFF
--- a/docs/source/users/infer_tutorial.rst
+++ b/docs/source/users/infer_tutorial.rst
@@ -175,7 +175,7 @@ prior and using it in multiple places :
     s2 = Sphere(n=n_prior, r=prior.BoundedGaussian(.5, .05, 0, np.inf),
                 center=prior.make_center_priors(data_holo))
 
-    spheres = Spheres([s1, s2, s3])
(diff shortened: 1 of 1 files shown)


rbnvrw:
I see the Travis build is failing, but I don't think this has anything to do with my documentation update.

tdimiduk:
I like what you have written, I am not sure if that is the right place to put it in the docs @barkls or @vnmanoharan do you want to comment on where this should go in the inference docs?

@barkls can you comment on the Travis issue? Are Travis builds failing right now for other reasons, or did something @rbnvrw do break the doctest build?

barkls:
I also like the clarifying documentation. I think the current location is fine. Eventually, we may have an "advanced techniques" section in the inference tutorial page, but we don't have one now so it makes sense to put it with the information on defining priors. 

The travis build failure is almost certainly unrelated. I haven't changed anything or run travis in a few months. It's possible that something on travis has changed in the meantime, or it may be related to #194.

PR #216, opened 2018-03-12
A few changes I have accumulated

barkls:
I think it makes sense to merge in changes as they happen, and not just when preparing for a release.


vnmanoharan:
Travis is failing on `holopy.scattering.tests.test_multisphere.test_cross_sections`

barkls:
Yes, This appears to be an instance of #194. Specific multisphere tests sometimes fail on Travis for PRs, but then they pass when merged in to the main branch. I don't understand it. 

All tests passed locally and I didn't adjust anything in multisphere, so I am confident it is safe to ignore the failure this time. I realize the underlying problem needs to be fixed though, since that's the whole purpose of running tests on Travis.

Issue #88, opened 2016-10-19
multisphere with >20 spheres


barkls:
@AnnieStephenson pointed out that there is currently an upper limit of 20 spheres (specified in scfodim.for file). Is there a physical or computational reason for this hard limit?

vnmanoharan:
The Multisphere model currently relies on Fortran 77 code. Fortran 77 doesn't do dynamic memory allocation for arrays.  So what one has to do is to specify the largest size for the array ahead of time, which can be done by setting a few parameters in one of the Fortran files.  We could increase the number of spheres, but that has an impact on memory utilization even for small numbers of spheres (since the arrays are allocated at the maximum size).

The longer-term solution is to switch to Mackowski's F90 MSTM code.  We have a [python wrapper] (https://github.com/manoharan-lab/mstm-wrapper) around this code, but it would need some work to make it compatible with holopy.  In particular, we would need to:

* modify the Fortran code to output the amplitude scattering matrix rather than the scattering matrix
* modify the code to work with F2py, so that we can call functions directly from holopy rather than going through a pipe or input/output from a file
* get permission from Mackowski to distribute the modified code under the terms of the GPLv3

barkls:
I guess the short term question then becomes: What is the memory load per sphere? It looks to me like it's at least around 10 MB just from skimming the code. We could do some memory usage tests to decide if 20 spheres is a reasonable limit for modern computers, or we can just leave it until the longer-term solution is implemented.

vnmanoharan:
I think we can leave it up to the user.  Anyone who needs a larger number of spheres can simply fork the repository, make the change to scsmfo.dim, and then pull new revisions from the main branch if they want to keep up with changes. There should probably be a note somewhere in the documentation, though. @tdimiduk and @AnnieStephenson, any comments?

tdimiduk:
We set the 20 sphere number back when memories were smaller so pushing it to 40 or 50 might be reasonable. I agree with @vnmanoharan that it makes sense to include instructions on changing the value in the docs, or even provide a script to change the value and recompile (that might actually be easier than describing it in enough detail for a non expert to do the change).

vnmanoharan:
OK, then I suggest doing a few tests and seeing how memory use changes as we increase the value.  Let's set an upper bound of say 1 GB of memory (so that users with 32-bit machines can still operate the code). @barkls, can you look at this?

tdimiduk:
[...]

Issue #142, opened 2016-11-30
HoloPy won't compile on Windows with Anaconda and gcc


vnmanoharan:
I installed the MinGW gfortran from anaconda.  Also installed Visual C++ Build Tools from Microsoft.  Tried to build with
`python setup.py build_ext --inplace --fcompiler=gfortran`
Result is that f2py works, but linking fails:
```
uts_scsmfo.o : error LNK2001: unresolved external symbol ___chkstk_ms
uts_scsmfo.o : error LNK2001: unresolved external symbol _gfortran_pow_i4_i4
uts_scsmfo.o : error LNK2001: unresolved external symbol _gfortran_pow_c8_i4
SBESJY.o : error LNK2001: unresolved external symbol _gfortran_st_write
SBESJY.o : error LNK2001: unresolved external symbol _gfortran_transfer_real_write
SBESJY.o : error LNK2001: unresolved external symbol _gfortran_st_write_done
```

Also tried to build with gfortran + MinGW C compiler (instead of MSVC):
`python setup.py build_ext --inplace --compiler=mingw32`
But this fails with
```
  File "setup.py", line 101, in <module>
    package=['HoloPy'])
  File "Anaconda3\lib\site-packages\numpy\distutils\core.py", line 166, in setup
    return old_setup(**new_attr)
  File "Anaconda3\lib\distutils\core.py", line 148, in setup
    dist.run_commands()
  File "Anaconda3\lib\distutils\dist.py", line 955, in run_commands
    self.run_command(cmd)
  File "Anaconda3\lib\distutils\dist.py", line 974, in run_command
    cmd_obj.run()
  File "Anaconda3\lib\site-packages\numpy\distutils\command\build_ext.py", line 117, in run
    force=self.force)
  File "Anaconda3\lib\site-packages\numpy\distutils\ccompiler.py", line 596, in new_compiler
    compiler = klass(None, dry_run, force)
  File "Anaconda3\lib\site-packages\numpy\distutils\mingw32ccompiler.py", line 58, in __init__
    dry_run, force)
  File "Anaconda3\lib\distutils\cygwinccompiler.py", line 157, in __init__
    self.dll_libraries = get_msvcr()
  File "Anaconda3\lib\distutils\cygwinccompiler.py", line 86, in get_msvcr
    raise ValueError("Unknown MS Compiler version %s " % msc_ver)
ValueError: Unknown MS Compiler version 1900
```
[...]

PR #220, opened 2018-04-05
Make HoloPy build on Windows with conda-forge

barkls:
conda-forge can now use [flang](https://github.com/flang-compiler/flang) to compile Fortran on Windows. This PR includes 3 changes to make HoloPy build in this environment:

1) Switch (revert) to using double precision axisymmetric T-matrix code. The flang compiler doesn't support quad-precision variables so we have to switch back to double precision, which I salvaged from an old PR.

2) Removal of mie_f/scfodim.for parameter file for multisphere since Fortran's ``include`` command wasn't finding it. There are several degrees of separation between us and what is actually doing the compile, since it goes through conda-forge's machinery and also f2py, so it's not easy to add a flag when calling the compiler to help find the file. I just pasted the file contents everywhere we were supposed to include it. Note that this file doesn't allow us to easily change parameters anyway, see #88 comments thread.

3) Let setup.py bypass compiling Tmatrix code. Our Makefile has gfortran hard-coded, but flang is a separate compiler. I can compile Tmatrix code with a ``bld.bat`` file independent of setup.py that will run only on conda-forge.


barkls:
The failing test on Travis is a Multisphere failure, see #194. All tests passed locally.

vnmanoharan:
Excellent!  Well done.  I wonder if we can put these changes in a different branch and tell conda-forge to compile the windows package from that branch.  I would like to keep quad precision for linux and mac, since it might be needed to handle axisymmetric objects with high aspect ratios.

barkls:
I'm not certain, but I expect building from two separate branches is possible with conda-forge. However, I'm not sure that is the best approach, since we would then have to merge all future changes into both branches. An alternative would be to distribute both versions of the Fortran code (approx. 180 kB each) and compile the appropriate version in each case.

vnmanoharan:
I don't think merging will be a big problem, since most of the edits are to files that don't get changed much. So merging changes to a windows-build branch from the master branch should rarely give conflicts.

barkls:
Ok I've changed this PR to merge into a new branch. It is worth clarifying in the docs that users trying to build from source on windows should clone master, and not windows-build (which now lacks a T-matrix Makefile - maybe I should just put it back?)...

tdimiduk:
Is there not a way to something with preprocessor directives and makefiles
to that will work on windows but also give the precision we want on linux?
Or by having a small fortran file that controls the precision that we build
against different versions on windows and linux? Or (this may be a really
bad idea) have a setup.py script which modifies the code when building on
windows to the needed precision before building? From what I have absorbed
of packaging wisdom from open source communities, I think if there is some
way to use the same codebase without significant code duplication, that
might be preferable.

vnmanoharan:
@barkls I think it's fine to leave the makefile out of windows-build. To advise users on which branch to use for building, I recommend modifying the README.md file, since that advice is specific to GitHub. 

We should also update the docs to describe building on Windows with gfortran.  Pinging @ralex0 here to help with that task.

vnmanoharan:
@tdimiduk The master branch compiles fine on any platform with gfortran, including windows.  The only issue is that conda-forge can't compile with gfortran, so the branch would only be used in conda-forge's automated builds of conda packages.  The master branch will remain usable and compile-able on all platforms.

barkls:
The docs currently do have instructions that @ralex0 put together for compiling on Windows with gfortran.

Are we set to merge? (Failing Travis is still due to #194)

tdimiduk:
Ah, that is probably reasonable then. There is value to having a single
codebase for all platforms (which I am comfortable calling conda-forge a
platform here), but you are probably right that it isn't worth complicating
things for that, people do release branches and backport things all the
time and this looks like it will be simpler than that.

vnmanoharan:
I will merge. @barkls, could you please file an issue saying that quad-precision compilation and includes don't work with flang? That will address @tdimiduk's point about flang not working as a platform with master.  We may be able to fix this in the future.  Also please file an update to #142, which I think is OK to close now.

PR #226, opened 2018-05-31
2colour

### LINKED ISSUE #158: Simultaneous inference from 2 hologram colours
This should be relatively straightforward with xarray - we will need to change wavelength from an attribute to a dimension for these images,  and propagate how to handle this object through all code. Scattering and inference calculations will need to know how to behave appropriately.

barkls:
This fixes #158, among other things


barkls:
The only failing test addresses multisphere cross section calculations, which were not changed in this PR. It seems to be a case of #194. 

I think the code is ready to merge and then release v3.2. Any objections @vnmanoharan?

vnmanoharan:
OK to merge with 3.2 branch, but before merging with master we should revisit issues filed for [3.2 milestone](https://github.com/manoharan-lab/holopy/milestone/4) and retarget or close, as appropriate.

barkls:
That's a good point. Only one of those issues (#158) is addressed here. We could call this 3.1.3 if we want to reflect that. However, I think it's worth releasing this as 3.2.0 and pushing the remaining issues to a future release for 2 reasons:
1) Adding support for multichannel holograms is a fairly significant feature addition.
2) This update fixes a major bug in inference calculations, so it would be nice to give the version a significant bump.

vnmanoharan:
See my comment above.

PR #240, opened 2018-06-28
convolution_propagation: make gradient_filter work, update docstrings and fix style

korigod:
1. `sqrt` function was earlier imported from `xarray.ufuncs`. The import was later removed (commit bbe4cf3) so now use of `gradient_filter` leads to exception.
2. `trans_func` docstring is outdated so I've updated it. Please advice me on the right `schema` arg description, I'm not sure I've written it the best way (the last commit).
3. Style fixes to comply with PEP8 and a bit more.


korigod:
> When we moved from HoloPy 2-->3 we removed our own Image and VectorGrid classes, in favour of representing holograms as xarray.DataArray objects, so this is what format schema should take.

Thanks, I've corrected it (in `propagate` docstring too).

> We would really appreciate it if you can add a doctest or unit test that runs this code as well.

I've added the test. I also have formatted the yamls to look pretty (as yaml should :)). Unfortunately, `core.tests.common.make_golds()` doesn't work so I have not fixed it to generate pretty yamls. At first it didn't work because there was a deprecation warning when `save` is invoked (the problem is in `xarray/backends/common.py`, not in `holopy`) and `core.tests.common` fails on warnings. I've commented out the string that fails on warnings and then `TypeError: 'complex' object is not iterable` is thrown by `yaml/representer.py`. So I just wrote `gold.yaml` by hand (without full version).

PR #247, opened 2018-11-13
Fitting rework

barkls:


barkls:
All unit-tests now passing. Doctests are failing, but this is expected since the user docs haven't been updated to reflect new API yet.

PR #250, opened 2018-11-30
Cleanmielens

briandleahy:
This PR is just for a few minor clarifications + speedups to the mielens code; there should be no functional difference.


briandleahy:
This fails Travis, but only from doctest issues. There are 2 failing unittests which are old, known issues that fail on the base 3.3 branch.

PR #253, opened 2019-02-14
3.3

### LINKED ISSUE #219: Coordinate image saving and displaying
Image brightness and contrast is auto-scaled when displaying images. This makes it difficult to compare two holograms, since their mapping will be different. We should have an option to pass in a scaling range, as when saving images.

barkls:
also fixes #219.


barkls:
unit tests are passing on Travis. Doctests are failing, but that's expected due to version inconsistencies and altered API not reflected in docs (yet).

barkls:
doctests still failing, unittests pass.

PR #259, opened 2019-05-10
3.3

### LINKED ISSUE #255: HoloPy dependencies
HoloPy has a lot of dependencies, with some new ones introduced in 3.3. Many of these (cma, emcee, h5, etc.) are only used in some parts of HoloPy and should be made optional, with intelligent error handling and test skipping to compensate when they are not present.

barkls:
fixes #255.

### DIFF
--- a/holopy/__init__.py
+++ b/holopy/__init__.py
@@ -31,10 +31,10 @@ images and calculations.
 """
 
 
-from . import core, scattering, fitting, inference
-from .core import (load, save, load_image, save_image, show, test_display,
+from holopy import core, scattering, fitting, inference
+from holopy.core import (load, save, load_image, save_image, show, test_display,
                                             detector_grid, detector_points)
-from .propagation import propagate
+from holopy.propagation import propagate
 
 __version__ = '3.3.0'
 __version_info__ = tuple([ int(num) for num in __version__.split('.')])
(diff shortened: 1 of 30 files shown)


briandleahy:
Thanks for taking the time to fix this @barkls . A few comments:

0. The tests are failing...
1. holopy.core.utils, line 172 -- "pass in pallel=None" -- parallel is mis-spelled :)
2. cmaes, line 136 -- it's probably best to try to import cmaes at the top of the file (per pep8). You could do something like this at the top:
```python
try:
    import cmaes
except ModuleNotFoundError
    _HAS_CMAES = False
```
...
```python
def run_cma(...):
    if not _HAS_CMAES:
        raise DependencyMissing(...)
```
The same thing applies to `emcee`, lines 117, and `multisphere`, lines 127
3. cmaes, line 173: It might be a little cleaner with tempfile.TemporaryDirectory as a context manager, see https://docs.python.org/3/library/tempfile.html#tempfile-examples. Thoughts?
4. Thanks for tightening up that code in `holopy.scattering.theory.dda` :+1: 
5. Can I ask you to remove the relative import in `hp.scattering.theory.mie_f.__init__.py`

Also, if you have the time, it might be worth fixing some pep8 items on these files. For instance, there is some trailing whitespace in the files. I don't know if it's best to fix this as part of this PR or as something more stand-alone.

barkls:
Tests are passing now, it's just doctests that cause Travis to fail. @briandleahy can you have another look when you get a chance and either merge or suggest other changes?

PR #260, opened 2019-05-15
cleancalcholo

briandleahy:
This set of commits just cleans up some of the code in scattering.theory.calculations.py, along with some other minor pep8 and style changes throughout some other files which touch calculations.py. There should be no changes in any functionality; just cosmetic changes.


holopy/scattering/calculations.py
```diff
@@ -39,24 +38,27 @@
     from holopy.scattering.theory.dda import DDA
 except:
     pass
+    # FIXME why is this a try-pass?
```
barkls: It doesn't need to be anymore. It should now be safe to do these imports even if the relevant scattering theories aren't compiled/installed since the error is now raised further downstream.

holopy/scattering/calculations.py
```diff
@@ -178,22 +199,15 @@ def calc_holo(schema, scatterer, medium_index=None, illum_wavelen=None,
...
         Calculated hologram from the given distribution of spheres
     """
-
-    # Essentially all the time is in theory._calc_field
-    scaling = dict(_expand_parameters({'alpha':scaling}.items()))  # 6 us
-    for key in scaling.keys():
-        if hasattr(scaling[key],'guess'):
-            scaling[key] = scaling[key].guess
-    scaling = _interpret_parameters(scaling)['alpha']  # 4 us
-    scaling = dict_to_array(schema, scaling)  # 754 ns
```
barkls: This code modifies scaling for the cases where it is either a fixed value or a `hp.Inference.prior.Prior object`, and can also be passed in as a dictionary or xarray for cases where there are multiple values of scaling - eg. 2 colours. Maybe this isn't tested (it should be), but you shouldn't delete unless you have confirmed it works in all of those cases.

holopy/scattering/theory/dda.py
```diff
@@ -36,11 +33,14 @@
...
-from holopy.scattering.theory.scatteringtheory import ScatteringTheory
-from holopy.scattering.scatterer import (Ellipsoid, Capsule, Cylinder,
-                                        Bisphere, Sphere, Scatterer, Spheroid)
+import numpy as np
+
 from holopy.core.utils import ensure_array
-from holopy.core.errors import DependencyMissing
+from holopy.scattering.scatterer import (
+    Ellipsoid, Capsule, Cylinder, Bisphere, Sphere, Scatterer, Spheroid)
+from holopy.scattering.errors import DependencyMissing
```
barkls: This should be switched back to importing from `holopy.core.errors` since I moved it when rejigging dependencies. This is causing unittests to fail on Travis.

holopy/scattering/theory/mie.py
```diff
@@ -144,10 +152,9 @@ def _raw_cross_sections(self, scatterer, medium_wavevec, medium_index, illum_pol
 
         where I_0 is the incident intensity.  See van de Hulst, p. 14.
         """
-        if isinstance(scatterer, Scatterers):
-            raise InvalidScatterer(scatterer,
-                                        "Use Multisphere to calculate " +
-                                        "radiometric quantities")
+        if not self._can_handle(scatterer):
```
barkls: This is too general for the corresponding error message. It refers specifically to Multisphere, but that is only valid for `Spheres` scatterers, and not arbitrary scatterers that cannot be handled by Mie.

holopy/scattering/calculations.py
briandleahy: Fixed with commit aa1066d

holopy/scattering/theory/dda.py
briandleahy: Oops, my bad :) I had un-commited changes on my machine. Should be fixed with commit f659172

holopy/scattering/calculations.py
briandleahy: Sorry I meant to mention this commit to you directly when I did this. I remembered there was some reason that the code was there, but I couldn't remember what it was.

Can I ask you to either add a test or a minimal-working-example for the use case that would break if the above code is removed? I can then quickly change my modification to make it work with that use case.

Thanks!

holopy/scattering/theory/mie.py
briandleahy: Per our discussion, I've changed this to
```python
if isinstance(scatterer, Spheres):
```

briandleahy: (in commit 2f67510)

holopy/scattering/calculations.py
barkls: It looks like the call to `dict_to_array` in line 193/202 isn't necessary since `from_parameters` puts the scatterer back in a dict anyway.

Here's a working example:
```python
import holopy as hp
from holopy.scattering import Sphere, calc_holo
from holopy.inference.prior import Uniform

detector = hp.detector_grid(5,1,extra_dims={'illumination':['red','green']})
fixed_scatterer = Sphere(r=0.5, n={'red':1.5,'green':1.6}, center=(2,2,2))
prior_scatterer = Sphere(r=0.5, n={'red':Uniform(1.5, 1.6), 'green':Uniform(1.5, 1.6)}, center=(2,2,2))
fixed_alpha = {'red':0.8,'green':0.9}
prior_alpha = {'red':Uniform(0.6,1),'green':Uniform(0.6,1)}
for alpha in [fixed_alpha, prior_alpha]:
    for scatterer in [fixed_scatterer, prior_scatterer]:
        calc_holo(detector, scatterer, scaling=alpha, illum_polarization=(0,1),
                  illum_wavelen={'red':0.66,'green':0.52}, medium_index=1.33)
```

briandleahy: Great, thanks! There are 2 things here -- the 2-wavelength alpha and the prior alpha. I added both of these as separate tests to ```test_2_color.py```. 

To make the tests pass I just added those 4 deleted lines back in, each of them is essential as of now. 

However, right now I still don't think the code is well-organized. The 4 lines I added back in are really at the wrong level of abstraction for the function. In addition, they do different things depending on what type of variable ```scaling``` is. I think for now,  though, I'm just going to leave it and move on to something else.

Thoughts?

barkls: Yes.

holopy/scattering/tests/test_2_color.py
```diff
@@ -21,36 +21,85 @@
...
+        scatterer = Sphere(
+            r=0.5, n={'red': 1.5, 'green': 1.6}, center=(2, 2, 2))
+        alpha = {'red': 0.8, 'green': 0.9}
+        result = calc_holo(
+            detector, scatterer, scaling=alpha, illum_polarization=(0, 1),
+            illum_wavelen={'red': 0.66, 'green': 0.52}, medium_index=1.33)
+        assert result is not None
+
+    @attr("fast")
+    def test_calc_holo_with_twocolor_priors(self):
```
barkls: Thanks for adding these tests!

barkls:
PR is ready to merge as far as I'm concerned.

PR #261, opened 2019-05-16
Bugfix in capsule model

blwh:
Small name error in _adda_capsule.
Changed mediumindex to medium_index


barkls:
Good catch! This function is actually removed (and the behaviour made more general) in the  [development branch](https://github.com/manoharan-lab/holopy/tree/3.3) though, so I'm closing the PR.

PR #263, opened 2019-05-24
Use f2py to compile T-Matrix code

ralex0:
The main change in this PR is the T-matrix scattering theory object is refactored. The Fortran is compiled on installation with f2py instead of using a makefile. The calculations are run by calling a module instead of using `subprocess`

One note is this uses the double precision version of the [t-matrix code](https://www.giss.nasa.gov/staff/mmishchenko/t_matrix.html). I couldn't get quad/extended precision to work on my either of my personal machines (Intel 64 bit Windows 10 and AMD 64-bit Fedora 30), don't think my processors have quad precision available.

Lastly I added a failing test: test_raw_fields_similar_to_mie. It's weird that this fails, because the analogous test for the scattering matrix passes.


ralex0:
`test_determine_default_theory_for_on_cylinder` and `test_determine_default_theory_for_on_spheroid` [fail on Travis-CI, but not on my local machines](https://travis-ci.org/manoharan-lab/holopy/builds/536984546)

vnmanoharan:
Thanks, Ron.  This is nice work.  A few notes:

1. The quad precision code doesn't require any quad precision hardware, as far as I know.  The quad precision executable that we were using previously (with `subprocess`) compiled file on my 64-bit intel machine. So it might be a problem with f2py.

2. You might check to see if increasing `NPN1` and `NPNG1` in `ampld.par.f` fixes the failing tests.  These parameters were previously set to 200 and 600 when we were using the quad precision code.  They should increase the precision of the T-matrix calculations.

3. When using the quad precision code, we had to modify the code to set the incident direction and polarization, and more importantly to loop over all the scattering angles that HoloPy needs.  It's more efficient to do this loop in Fortran because the calculation of the T-matrix needs to be done only once.  If you made similar changes to the Fortran code, it would be good to document them in the .f file.

ralex0:
@vnmanoharan 

The quad precision may very well be a problem with f2py. 

I figured out why the fields were off, I was comparing a near field point and including the radial dependence in the mie fields. I changed the test to use mie fields without radial dependence and now it passes.

I added a header and readme to the main Fortran file. Instead of editing the original source, I wrote subroutine in a different file that loops over the angles. By the way, what's the reason for restricting our calculations to use (1, 0) polarization? In this implementation, the incident polarization is passed to the Fortran subroutine as an argument.

vnmanoharan:
I don't know why the original polarization was set to [1,0], but it could be due to a difference in coordinate systems between HoloPy and the code.  We'll want to double check this.

Regarding the loop: the amp_scat_matrix subroutine first calculates the T-matrix, then uses the T-matrix to calculate the amplitude scattering matrix.  Once the T-matrix calculation is finished, it can be used to calculate the amplitude scattering matrix for any angle or for any orientation.  But if you loop over amp_scat_matrix, you end up calculating the same T-matrix over and over again.  Ideally, we calculate the T-matrix once and then use it to calculate the amplitude scattering matrix at each angle.  That speeds things up by a lot.

In the original code, you'll see that the loop happens inside amp_scat_matrix, after the T-matrix is calculated.  Can you implement that in this pull request?  It will involve editing the original Fortran and modifying the header.

ralex0:
Great suggestion, Vinny. I modified the loop so that it only calculates the T_Matrix once per call to `amp_scat_matrix`. `test_tmatrix.py` used to take about 20s to on my machine, and now it takes only 0.5s!

vnmanoharan:
Looks good.  I think it's almost ready to merge, though there's one minor point to resolve: it looks like MAXITER isn't ever used in ampld.lp.f.  If it's not used, you could undo the changes to the file.  If it is used (for example, it's an output parameter), you should add a header to the file listing the modifications and noting that it was modified with permission.

ralex0:
Okay, that commit is pushed. 

I also forgot to mention this PR also has a side effect of fixing the Travis CI build.

vnmanoharan:
Looks good to merge to me.

barkls:
Nicely done Ron! Thanks for also investigating the Travis doctest issue.

Did you confirm the tmatrix vs DDA tests pass? Travis doesn't have adda and skips those tests so it's important that somebody confirmed they pass.

ralex0:
The DDA test passes on both my local machines at least.

briandleahy:
I just want to say  :heart_eyes: on these commits, esp. on the additional tests and the 40x speedup, not to mention the f2py code rather than the manual build. :1st_place_medal:

Regarding @vnmanoharan 's comment
 > I don't know why the original polarization was set to [1,0], but it could be due to a difference in coordinate systems between HoloPy and the code. We'll want to double check this.

Can we figure this out and encode this knowledge into a unit test?

PR #270, opened 2019-06-07
Remove is_none utility function

rgov:
The comment for `is_none` says:

> Check if something is None.  
> This can't be done with a simple is check anymore because numpy decided that		
    array is None should do an element wise comparison.

I haven't been able to confirm this behavior of Numpy. The `is` operator in Python is not overloadable and always tests if its operands are the same object. I ran some tests in a Python 3 REPL and did not see the described behavior.

This change reverts back to the canonical `is` and `is not` operators. This should be a small performance (and readability) gain.


barkls:
I don't know the history of this function. @vnmanoharan or @tdimiduk do you know if it is safe to remove?

rgov:
Here is the commit that added it: cb577770dcb67251888e3f3d84aa92245a022a09

I looked through the Numpy 1.9.0 release notes and do not see anything about the `is` operator. I do see that there was a change to the equality operator `==` in Numpy 1.4.0 but this is from 2009...

rgov:
Tests passed, docs failed to generate.

vnmanoharan:
@barkls At the time, we didn't know how to interpret the word "comparison" in the FutureWarning from numpy (`comparison to None will result in an elementwise comparison in the future`).  We thought it might apply to both `is` and `==` but now we know that it only applies to `==`.  This is safe to merge, assuming build failure is unrelated.

barkls:
confirmed failing CI is due to doctests only, as fixed in #263.

tdimiduk:
That agrees with my memory. I implemented that is_none a long time ago to
get around a warning. If things have settled out in a way that we can use
is without a warning and get the correct comparison, then I am +1 on
getting rid of this bit of cruft.

PR #271, opened 2019-06-08
Fix reference to refimg in load_average

rgov:


barkls:
Thanks for catching this!

PR #273, opened 2019-06-18
check for corner case when j0 = 0 in SBESJY, added unit test that fai…

### LINKED ISSUE #249: Bug in SBESJY.F
The algorithm used to calculate spherical Bessel functions in SBESJY.F sometimes fails near zeros of j_0(x). See example:

```
eps = 1e-12
d = hp.core.detector_grid(shape=10, spacing=1)
d = hp.core.update_metadata(d,illum_wavelen=1, medium_index=1, illum_polarization=(1,0))

s_exact = hp.scattering.Sphere(r=1,n=1.1,center=(5,5,5))
s_close = hp.scattering.Sphere(r=1,n=1.1,center=(5,5,5-eps))

h_exact = hp.scattering.calc_field(d,s_exact)
h_close = hp.scattering.calc_field(d,s_close)

hp.show(h_exact)
hp.show(h_close)
```
The best solution is probably to use scipy's new functions 'spherical_yn` and `spherical_jn`, which do not have this issue and function for large x (unlike the old scipy function `sph_jnyn`). However, there are a couple barriers to their implementation.
1) They are much slower than SBESJY, mostly because they don't return all orders of j_n and y_n so the function needs to be explicitly called for each n. Since the functions are calculated by a recurrence relation, this results in many repeated calculations behind the scenes, especially when many values of n are required. 
2) SBESJY is called from within fortran code, so there would need to be some reorganization to either calculate spherical Bessel functions from outside with python and pass them in, or call the scipy functions from within fortran.
3) @briandleahy pointed out to me that since this only occurs within ~10^-14 of some zeros, there is nearly zero probability of this bug ever being relevant by chance alone. It is only relevant when using simple test cases like the one shown above. 

Another possible solution is to modify the code in SBESJY to fix the bug. Basically, the un-normalized j_n(x) are calculated by downward recurrence. Then j_0(x) is calculated explicitly, and compared to the value from recurrence. All j_n(x) are scaled accordingly. This fails sometimes when j_0(x) is very close to zero. In those cases, it makes sense to calculate j_1(x) explicitly and rescale all j_n(x) by that instead.

jeromefung:
Fixes #249.

SBJESY.F now checks for the corner case when j_0(x) = 0 (within machine precision). Instead of trying to re-scale the results of downward recursion with the tiny j_0 in that case, re-scale using j_1. 

In the attached Jupyter notebook, I successfully tested the updated SBESJY.F against `scipy.special.spherical_jn`. I had to `f2py` SBESJY.F to allow it to be called directly from Python; the notebook won't work with HoloPy built in the usual way.

I added a unit test to `test_mie.py` that failed prior to updating SBESJY.F but now passes. The issue arose in calculating the Mie scattered field when kr = n*pi, where n >=1 is an integer.

[test_fixed_sbesjy.pdf](https://github.com/manoharan-lab/holopy/files/3303483/test_fixed_sbesjy.pdf)

### DIFF
--- a/holopy/fitting/tests/test_minimizer.py
+++ b/holopy/fitting/tests/test_minimizer.py
@@ -99,4 +99,4 @@ def test_iter_limit():
     model = Model(par_s, calc_holo, 1.33, .66, illum_polarization=(1, 0), alpha = Parameter(.6, [.1, 1]))
     warnings.simplefilter("always")
     result = fit(model, holo, minimizer = Nmpfit(maxiter=2))
-    assert_obj_close(gold_fit_dict,result.parameters,rtol=1e-6)
+    assert_obj_close(gold_fit_dict,result.parameters,rtol=1e-5)
--- a/holopy/scattering/tests/test_mie.py
+++ b/holopy/scattering/tests/test_mie.py
@@ -360,3 +360,20 @@ def test_raw_fields():
   (0.003816460764514863-0.0015982360934887314j),
   (0.0012772696647997395-0.0039342215472070105j),
   (-0.0021320123934202356-0.0035427449839031066j)]])
+
+
+@attr('medium')
+def test_j0_roots():
+    # Checks for misbehavior when j_0(x) = 0
+    eps = 1e-12
+    d = detector_grid(shape=8, spacing=1)
+    d = update_metadata(d,illum_wavelen=1, medium_index=1,
+                                illum_polarization=(1,0))
+
+    s_exact = Sphere(r=1,n=1.1,center=(4,4,5)) # causes kr = 0 near center
+    s_close = Sphere(r=1,n=1.1,center=(4,4,5-eps))
+
+    h_exact = calc_field(d,s_exact)
+    h_close = calc_field(d,s_close)
+
+    np.testing.assert_allclose(h_exact, h_close)
--- a/holopy/scattering/third_party/SBESJY.F
+++ b/holopy/scattering/third_party/SBESJY.F
@@ -98,15 +98,34 @@ C------ CALCULATE THE L=0 SPHERICAL BESSEL FUNCTIONS DIRECTLY
         JP(0)   = -Y(0) - XINV * J(0)
         YP(0)   =  J(0) - XINV * Y(0)
         IF (LMAX .GT. 0) THEN
-                   OMEGA  =  J(0) / DEN
-                      SL  = ZERO
-            DO 40 L = 1 , LMAX
+           IF (J(0) .GT. DSQRT(ACCUR)) THEN ! usual case, nonzero j(0) 
+                OMEGA  =  J(0) / DEN
+                SL  = ZERO
+                DO 40 L = 1 , LMAX
                     J (L) = OMEGA * J (L)
                     JP(L) = OMEGA * JP(L)
                     Y (L) = SL * Y(L-1)   -   YP(L-1)
                       SL  = SL + XINV
                     YP(L) = Y(L-1)  -  (SL + XINV) * Y(L)
-   40       CONTINUE
+ 40              CONTINUE
+              ELSE              ! j(0) = 0, normalize with j(1) instead
+                 DEN = J(1)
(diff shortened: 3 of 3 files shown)


vnmanoharan:
@jeromefung, any idea how much of a performance penalty this change will involve?  We calculate a huge number of Bessel functions per hologram, and it seems that the check has to be run each time we call the function.  If this incurs a small penalty (a few percent), it won't matter.  But would another possibility be simply to check if n is an integer?  That would only have to be done once per hologram.

jeromefung:
Good question. I'll run some tests with a profiler. I'll check both direct calls to SBESJY (after f2py'ing it) as well as calculating holograms, which is the context we probably care more about.

I don't think a check-once-per-hologram approach could work because kr (the non-dimensionalized distance from the origin of the scattered wave expansion to the detector location) is different for every point on the detector.

jeromefung:
I used `%timeit` to compare the original and updated versions of SBESJY. The differences between the two versions are insigificant.

When calculating `sbesjy(243.2, 40)` (not a multiple of pi), I got 3.36 microseconds +/- 11.5 ns for the old code and 3.3 microseconds +/- 24 ns for the new code.

When calculating `sbesjy(100*np.pi, 40)` (which is handled differently by the new code), I got 4.2 microseconds +/- 68.3 ns for the old code and 4.11 microseconds +/- 29.9 ns for the new code.

I also tried calculating a 200 x 200 hologram:

```python
sphere = Sphere(n = 1.59, r = .5, center = (9, 12, 10))
medium_index = 1.33
illum_wavelen = 0.660
illum_polarization = (1,0)
detector = hp_new.detector_grid(shape = 200, spacing = .1)
```

With the old code, `calc_holo()` took 371 +/- 8 ms, and with the new code, it took 371 +/- 4 ms. 

Hence, I think the performance penalty of checking for the corner case is negligible.

vnmanoharan:
Thanks @jeromefung.  @barkls, please have a look.  I think this is good to merge.

barkls:
close and open PR to trigger travis build.

barkls:
This change seems to have caused one of our tests to fail:

```
======================================================================
FAIL: test_minimizer.test_iter_limit
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/miniconda3/envs/testenv/lib/python3.6/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/build/manoharan-lab/holopy/holopy/inference/tests/test_minimizer.py", line 84, in test_iter_limit
    assert_obj_close(gold_fit_dict,result.parameters,rtol=1e-6)
  File "/home/travis/build/manoharan-lab/holopy/holopy/core/tests/common.py", line 87, in assert_obj_close
    context='{0}[{1}]'.format(context, key))
  File "/home/travis/build/manoharan-lab/holopy/holopy/core/tests/common.py", line 64, in assert_obj_close
    assert_allclose(actual, desired, rtol = rtol, atol = atol, err_msg=context)
  File "/home/travis/miniconda3/envs/testenv/lib/python3.6/site-packages/numpy/testing/nose_tools/utils.py", line 1396, in assert_allclose
    verbose=verbose, header=header, equal_nan=equal_nan)
  File "/home/travis/miniconda3/envs/testenv/lib/python3.6/site-packages/numpy/testing/nose_tools/utils.py", line 779, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Not equal to tolerance rtol=1e-06, atol=0
tested_object[0:r]
(mismatch 100.0%)
 x: array(0.524805)
 y: array(0.524804)
-------------------- >> begin captured stdout << ---------------------
Iter       1    CHI-SQUARE =  2.256575191  DOF =  9992
   0:r = 1  
   0:center.0 = 1  
   0:center.1 = 1  
   0:center.2 = 1  
   1:r = 1  
   1:center.0 = 1  
   1:center.1 = 1  
   1:center.2 = 1  
   alpha = 1  
--------------------- >> end captured stdout << ----------------------
```

I think the tolerance on that test can be reduced to make it pass, since you've already confirmed the change doesn't substantially affect calculated values.

Issue #267, opened 2019-06-06
Reducing memory usage during image load


rgov:
In troubleshooting high memory usage of a process running HoloPy, I noticed that loading a grayscale TIF file takes approximately 200% more memory than necessary, and the memory usage does not subside quickly.

1. `load_image` calls `data_grid` and passes in the image data, which is say ~120 MB of grayscale values represented as doubles.

2. `data_grid` executes `arr=np.array([arr])` which copies the array.

3. `data_grid` later calls `update_metadata` which directly calls `a.copy()` and returns the copy.

Theoretically these shouldn't consume memory over time because of Python's reference counting, but I do not see the memory subsiding.

For (2), I found that `arr = np.expand_dims(arr, axis=0)` produces the same array shape without the overhead.

For (3) the copy is probably not necessary at load time and could be skipped.

---

Additionally, during a call to `load_average`, there is a large spike in memory usage. At the peak of computation it requires 600 MB of RAM *per* input TIF file (4864 × 3232 pixels, grayscale).

1. I believe `clean_concat` copies all of the image data into a single array.

2. `accumulator.mean('images')` causes a temporary spike that partially subsides. Note that this calculation can be saved and re-used instead of performed twice. The resulting object occupies about the same size as a single image.

3. The calculation of `noise_sd` also causes a spike that partially subsides.

4. `copy_metadata` (optionally) and then `update_metadata` both make copies of their inputs.

barkls:
Thank you for finding these.

For your first set of comments on load_image, I agree your solution to point 2 is better. There might be some edge cases where the copy() in point 3 is necessary - I'll take a look.

For your second set of comments on load_average, this is an issue that is actually already on our radar, but not posted on github. @ralex0 are you looking into this?

ralex0:
Yeah, I am actually already working on the load_average performance. @briandleahy noticed this same issue before. 

@rgov, what branch are you using? Your clone of the master?

rgov:
I am using HoloPy installed from Conda Forge. I don’t know the version at the moment.

rgov:
As I understand it:

* `copy_metadata` returns a copy of an input array, attaching metadata copied from a second input array (potentially the same one)

* `update_metadata` returns a copy of an input array, replacing certain metadata fields with the given input parameters.

In many cases in the code, I see patterns like this that result in unnecessary copies:

```python
# makes a copy of data and then releases the original, so that only the copy remains
[...]

PR #276, opened 2019-06-28
Reduce memory usage by computing load_average online

ralex0:
These commits modify load_average. It now:
1. Loads one image at a time to save memory
2. Computes the mean and standard deviation in one pass using Welford's algorithm. 

This should alleviate some of the performance issues mentioned by @rgov in #267

Also included is a test to check performance, but there is room to change the threshold that I chose.


holopy/core/tests/test_io.py
```diff
@@ -33,12 +33,17 @@
...
 from holopy.core import load, save, load_image, save_image
 from holopy.core.errors import NoMetadata
+from holopy.core.io import load_average, get_example_data_path
 from holopy.core.process import normalize
 from holopy.core.metadata import get_spacing, copy_metadata
 from holopy.core.holopy_object import Serializable
 from holopy.core.tests.common import (
     assert_obj_close, assert_read_matches_write, get_example_data)
 
+import memory_profiler
```
barkls: Tests are currently failing since travis doesn't have this installed. Add it to `.travis.yml`.

briandleahy: Hmm... I disagree. I don't think we need ```memtest``` as a dependency for 1 unit test. I would rather do one of the following:
1.  Skip the unittest if memtest isn't installed.
2.  Remove the unit test which checks memory.

holopy/core/io/io.py
```diff
@@ -413,31 +413,57 @@ def load_average(filepath, refimg=None, spacing=None, medium_index=None, illum_w
...
 
     # overwrite metadata from refimg with provided values
-    return update_metadata(accumulator, medium_index, illum_wavelen, illum_polarization, normals, noise_sd)
+    return update_metadata(mean_image, medium_index, illum_wavelen, illum_polarization, normals, noise_sd)
+
+class Accumulator:
+    def __init__(self):
+        self.n = 0
+        self.m = None
+        self.s = None
```
briandleahy: I think this accumulator class is great, but can I just ask for longer variable names, e.g. ```self._running_mean``` instead of ```self.m```? Also, ```self.n, self.m, self.s``` are "private", right?

holopy/core/tests/test_io.py
```diff
@@ -178,9 +183,26 @@ def __init__(self, a):
...
+
+
+def _load_raw_example_data():
+    imagepath = get_example_data_path('image01.jpg')
+    return load_image(imagepath, **IMAGE01_METADATA)
+
+def _load_example_data_backgrounds():
+    bgpath = get_example_data_path(['bg01.jpg', 'bg02.jpg', 'bg03.jpg'])
+    return [load_image(path, **IMAGE01_METADATA) for path in bgpath]
+
```
briandleahy: In keeping with TDD, can we get some unittests on the ```Accumulator``` class?

holopy/core/tests/test_io.py
barkls: I wasn't suggesting we make it a dependency, just that we get travis to install it before running tests. I agree removing the test would be ok, since it's not testing functionality in the usual way.

holopy/core/io/io.py
briandleahy: Fixed by commits 42da7e3 and c1c08d4

holopy/core/tests/test_io.py
briandleahy: Fixed by c1c08d4

holopy/core/tests/test_io.py
ralex0: This is fixed by 02a7155

holopy/core/io/io.py
```diff
@@ -413,31 +413,57 @@ def load_average(filepath, refimg=None, spacing=None, medium_index=None, illum_w
...
 
+    # calculate the average
+    accumulator = Accumulator()
+    for path in filepath:
+        accumulator.push(load_image(path, spacing, channel=channel))
+    mean_image = accumulator.mean()
+
+    # calculate average noise from image
+    if noise_sd is None and len(filepath) > 1:
+        noise_sd = ensure_array(accumulator.std())
```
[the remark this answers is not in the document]
ralex0: Fixed by 42da7e3 and c1c08d4

holopy/core/io/io.py
```diff
@@ -413,31 +413,57 @@ def load_average(filepath, refimg=None, spacing=None, medium_index=None, illum_w
...
+
+        if self.n == 1:
+            self.m = x
+            self.s = x*0
+        else:
+            self.s += (x - self.m) * (x - (self.m + (x - self.m) / self.n))
+            self.m += (x - self.m) / self.n
+
+    def mean(self):
+        return self.m if self.n else 0.0
```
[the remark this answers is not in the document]
ralex0: Fixed by 42da7e3

holopy/core/io/io.py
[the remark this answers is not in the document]
ralex0: Fixed by 08a5672 then refactored and fixed by c1c08d4

ralex0:
Okay I think I have addressed al the comments left here. Thanks for the review. Let me know if there is anything else we should do before merging.

holopy/core/io/io.py
```diff
@@ -413,31 +414,69 @@ def load_average(filepath, refimg=None, spacing=None, medium_index=None, illum_w
...
+    def mean(self):
+        return self._running_mean if self._running_mean is not None else 0.0
+
+    def cv(self):
+        """ The coefficient of variation
+        """
+        if self._n == 0:
+            return None
+        else:
+            return np.mean(np.array(self._std() / self.mean()))
```
barkls: This won't work for multicolour images. Previously, the mean was taken only over dimensions {x,y,z} but now it's been replaced with np.mean which averages over all dimensions.

ralex0: Is the color axis dimesion 0 or -1?

barkls:
I don't know. the point of using xarray is that it doesn't matter since the
dimensions have names. Really, it should be able to handle either but I
expect all holograms are being created by ``data_grid`` so they should all
be the same.

ralex0:
Okay thanks, see my edit above about test data. If you don't have any I suppose I can make some from the existing data files, but they won't be really two color holograms.

holopy/core/io/io.py
briandleahy: Good catch @barkls . @barkls since you are the one who primarily uses the 2-color hologram, can I ask you to write a failing test? I'm worried that if @ralex0 writes it, the test will be incomplete, and there is no point in having a test that doesn't actually catch the desired use case.

barkls:
I can't push to your branch, so I'm just going to put the test here and you can copy it in. It should be part of the TestAccumulator class

```python
def test_2_colour_noise_sd(self):
    image = load_average('./2_wl_bg/', image_glob='*.jpg', spacing=1, channel=[0,1])
    gold_noise = [0.06864433355667054, 0.04913377621162473]
    noise = [image.noise_sd.loc[colour].item() for colour in ['green', 'red']]
    self.assertTrue(np.allclose(gold_noise, calculated_noise)
```
Here are the 4 images used in the test:





Let me know if you have any questions or if there's anything else I can help with!

holopy/core/io/io.py
ralex0: New test is in and passing now.

barkls:
Hi @ralex0 is this ready for another code review or are you still working on it?

ralex0:
It should be all set!

holopy/core/io/io.py
```diff
@@ -410,34 +411,82 @@ def load_average(filepath, refimg=None, spacing=None, medium_index=None, illum_w
         spacing = get_spacing(refimg)
 
     # read colour channels from refimg
+    channel_dict = {'0': 'red', '1': 'green', '2': 'blue'}
     if channel is None and refimg is not None and illumination in refimg.dims:
         channel = [i for i, col in enumerate(['red','green','blue']) if col in refimg[illumination].values]
```
barkls: The enumerate here should be switched to iterating over `channel_dict.items()` since you've defined that dictionary. I don't think order matters here, and dictionaries are ordered as of python 3.6 anyway.

barkls: Note this is a minor concern and I'm happy to go ahead and merge if you are sick of working on this and don't want to fix it.

PR #279, opened 2019-07-11
Minor fixes, including allowing inference of noise_sd

barkls:
This change (underscore -> colon in parameter parsing) might cause errors when running new methods on previously saved multicolour holopy results objects, but they should be loaded into HoloPy just fine and any bleeding-edge users can make the necessary adjustments themselves.


holopy/core/holopy_object.py
```diff
@@ -83,8 +80,15 @@ def _iteritems(self):
...
-
+        tag = '!{0}'.format(data.__class__.__name__)
+        value = []
+        # heavily inspired by the source of PyYAML's represent_mapping
+        node = yaml.nodes.MappingNode(tag, value)
+        for key, item in data._iteritems():
+            node_key = dumper.represent_data(key)
+            node_value = dumper.represent_data(item)
+            value.append((node_key, node_value))
+        return node
```
briandleahy: This is more of a question than a suggestion -- are there tests for these? If not it would be AWESOME to add some in, as if you TDD'd this with the refactor. Given that the yaml was causing some problems before, a few tests here would really help. 

If there are already tests, then ignore this please :)

And thanks for putting the ```ordered_dump``` function in the class method, which is the only place it's used.

holopy/inference/tests/test_prior.py
```diff
@@ -38,6 +38,7 @@ def test_uniform():
     assert_equal(u.guess, 1.5)
     assert_equal(u.interval, 1)
     assert_raises(ParameterSpecificationError, Uniform, 4, 6, 7)
+    assert_raises(ParameterSpecificationError, Uniform, 4, 4)
```
briandleahy: Thanks for adding this test!

holopy/scattering/tests/test_scatterer.py
```diff
@@ -117,16 +117,31 @@ def test_Sphere_parameters():
...
+        self.assertTrue(expansion == self.expanded)
+
+    @attr("fast")
+    def test_interpret_parameters(self):
+        dict_f = {'Left': {'H': 11, 'He': 12, 'Li': 13},
+                  'Right': {'H': 14, 'He': 15, 'Li': 16}}
+        simple_compressed = {'a':0, 'b':[0.5, 1, 2], 'c':{'c1':3, 'c2':4},
+                             'd':5+6j, 'e':{'e1':7+8j, 'e2':[9,10]},'f':dict_f}
+        compression = _interpret_parameters(self.expanded)
+        self.assertTrue(compression == simple_compressed)
```
briandleahy: Thanks for splitting this test into two separate tests that each only test one thing, and thanks for adding the ```@attr('fast')``` tag!

holopy/core/holopy_object.py
barkls: There is one minimal test of this to make sure the code runs. It's `test_yaml_output_of_serializable` found in test_io.py. I added to it a bit to make sure it has the correct behaviour about which attributes to include, but this still isn't sufficiently tested.

Issue #280, opened 2019-07-12
Insufficient testing


barkls:
We identified a few of the worst files and assigned them to various developers to fix:

Goal is to write tests if it's straightforward, or flag for a refactor with tdd if it's a mess, so as to avoid spending time/effort writing tests for code that really needs to be rewritten

@barkls 
hp.scattering.scatterer.scatterer.py (and scatterers.py, sphere.py, spheres.py...)
hp.core.utils.py
hp.inference.prior.py, hp.inference.model.py

@briandleahy 
hp.core.metadata.py
hp.core.io.io.py (with @ralex0)
hp.scattering.theory.scatteringtheory.py
hp.scattering.theory.mielens.py
hp.scattering.theory.mielenstheory.py

@ralex0 
hp.core.io.vis.py 
hp.core.io.io.py (with @briandleahy)
hp.scattering.calculations.py
hp.scattering.errors.py

briandleahy:
Adding more theory tests in #282

barkls:
added tests to prior.py in #287

briandleahy:
Testing metadata in #286

barkls:
io and vis still aren't sufficiently tested but they should really be refactored since the code is in pretty bad shape. We made good progress on everything else!

PR #281, opened 2019-07-18
Improve testing of utils.py

barkls:
This is my first contribution to #280.

### DIFF
--- a/holopy/core/tests/test_utils_math.py
+++ b/holopy/core/tests/test_utils_math.py
@@ -249,34 +249,33 @@ def test_rotation_matrix_degrees():
 #test utils
 
 class TestEnsureArray(unittest.TestCase):
-    np_array = np.array([1])
-    xr_array = xr.DataArray([2], dims='a', coords={'a':['b']})
-
     @attr("fast")
     def test_None_is_unchanged(self):
         self.assertTrue(ensure_array(None) is None)
 
     @attr("fast")
     def test_xarray_is_unchanged(self):
-        self.assertTrue(self.xr_array.equals(ensure_array(self.xr_array)))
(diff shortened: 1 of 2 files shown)


holopy/core/tests/test_utils_math.py
```diff
@@ -238,41 +239,167 @@ def test_rotation_matrix_degrees():
...
+    def test_None_is_unchanged(self):
+        self.assertTrue(ensure_array(None) is None)
+
+    @attr("fast")
+    def test_xarray_is_unchanged(self):
+        self.assertTrue(self.xr_array.equals(ensure_array(self.xr_array)))
+
+    @attr("fast")
+    def test_listlike(self):
+        self.assertTrue(ensure_array([1]) == self.np_array)
```
briandleahy: FYI the ```unittest.TestCase``` methods have an ```assertEqual``` method, which gives a nicer error message when it fails. (It says what two things are not equal, rather than just this is False).

holopy/core/tests/test_utils_math.py
```diff
@@ -238,41 +239,167 @@ def test_rotation_matrix_degrees():
...
+
+    @attr("fast")
+    def test_nothing_to_repeat(self):
+        repeated = repeat_sing_dims(self.input_dict, ['x', 'y'])
+        self.assertTrue(repeated == self.input_dict)
+
+class TestChoosePool(unittest.TestCase):
+    @attr("fast")
+    def test_custom_pool(self):
+        class DummyPool():
```
briandleahy: It might be nice to move the ```DummyPool``` outside this test case, so that multiple tests can use it easily if desired.

PR #287, opened 2019-08-01
Test prior.py

barkls:
I've written tests that cover all of prior.py. Tried a few new (for me) things here so let me know what you think.


holopy/inference/tests/test_prior.py
```diff
@@ -17,60 +17,320 @@
...
+        self.assertTrue(np.all(samples < bounds[1]))
+
+    @attr("medium")
+    def test_sample_distribution(self):
+        n_samples = 100000
+        n_quantiles = 10
+        bounds = np.random.rand(2) + np.array([0,1])
+        u = Uniform(*bounds)
+        samples = u.sample(n_samples)
+        quantiles = np.quantile(samples, np.linspace(0, 1, n_quantiles))
```
briandleahy: This line (104) fails on Travis because ```np.quantile```wasn't implemented until numpy version 1.15. I'd suggest using ```np.percentile```, although then you have to change it to
```python
         quantiles = np.percentile(samples, np.linspace(0, 100, n_quantiles))
```

Also, I have some other comments on the test:
* I would set a random seed, since a probabilistic test should fail with a small probability, you don't want to run the test suite 1000 times and have it fail once and be super confused as to why that's the case.
* 100,000 samples is a lot of samples. I'd make the number much fewer (e.g. 10,000 or 1,000) and the corresponding tolerance much smaller.
* Since we're physicists who like inference here, I'll suggest something a little more technical. By testing quantiles, you're only testing the distribution's value at 10 points. You can instead test the distribution at *every* value using what's called a [Kolmogorov-Smirnov test](https://en.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test) -- there are other tests for equality of distribution, but I think this is the best test for the uniform distribution case. An advantage of a KS test here is that not only does it test for equality of distribution everywhere (rather than at 10 points), it is also *much* more efficient and very easy to calculate P-values for. Fortunately there is a KS test implemented in scipy (```scipy.stats.kstest```) which calculates P-values etc, so this is a easy change. I'm happy to chat if you have questions.
* Also, since we're still all physicists here, I'd rather see ```atol``` and ```rtol``` to be something you'd expect from the statistical test, rather than an arbitrary number (e.g. a P-value or something).

Finally, at this stage I'd rather have many B+ tests than a few A+ tests, so I'm happy with only fixes that partially address these comments.

holopy/core/tests/test_utils_math.py
```diff
@@ -403,9 +404,14 @@ def test_schwimmbad_multipool(self):
...
         if MPIPool.enabled():
+            # mpi should work
             mpi_pool = choose_pool('mpi')
             self.assertTrue(isinstance(mpi_pool, MPIPool))
+        elif find_spec('mpi4py') is None:
+            # mpi4py not installed
+            self.assertRaises(ImportError, choose_pool, 'mpi')
         else:
+            # mpi4py installed but only one process available
             self.assertRaises(ValueError, choose_pool, 'mpi')
```
briandleahy: Thanks for fixing these. I'll try to run this on my not-mpi-installed machine and see if it works now.

holopy/inference/prior.py
```diff
@@ -28,6 +29,10 @@
 EPS = 1e-6
 
 class Prior(HoloPyObject):
+
+    def __init__(self):
+        raise NotImplementedError("Use subclass with a defined probability"
+                                  "distribution method prob and/or lnprob.")
```
briandleahy: :+1:

holopy/inference/prior.py
```diff
@@ -285,13 +307,19 @@ def get_val(val):
                 return val
         return get_val(self.real) + 1.0j * get_val(self.imag)
 
+    def __add__(self, value):
+        if isinstance(self, ComplexPrior) and isinstance(value, ComplexPrior):
+            return self.__addadd__(value)
```
briandleahy: Is ```__addadd__``` a dunder method? I'm super confused...

holopy/inference/tests/test_prior.py
```diff
@@ -17,60 +17,320 @@
...
+
+class TestBasics(unittest.TestCase):
+    @attr("fast")
+    def test_cannot_instantiate_baseclass(self):
+        self.assertRaises(NotImplementedError, Prior)
+
+class TestUniform(unittest.TestCase):
+    @attr("fast")
+    def test_construction(self):
+        parameters = {'lower_bound':1, 'upper_bound':3, 'guess':2, 'name':'a'}
```
briandleahy: pep8!!!!

briandleahy: In general it would be good to run these files through a pep8 checker. I like [pep8online.com](www.pep8online.com). It's pretty good overall, but I'm noticing a few mistakes (no space after colons and commas in a few places, for example).

holopy/inference/tests/test_prior.py
```diff
@@ -17,60 +17,320 @@
...
+        u = Uniform(*bounds)
+        self.assertRaises(AttributeError, setattr, u, 'interval', 2)
+
+    @attr("fast")
+    def test_prob(self):
+        bounds = np.random.rand(2) + np.array([0,1])
+        u = Uniform(*bounds)
+        self.assertEqual(u.prob(0), 0)
+        self.assertEqual(u.prob(2), 0)
+        self.assertEqual(u.prob(1), 1/np.diff(bounds))
```
briandleahy: Hmmm... I'm surprised this line doesn't have roundoff errors. I imagine that it's because you're calling the exact same operation in setting up the ```Uniform``` prior. But it might be worthwhile to change this to an
```python
self.assertAlmostEqual(u.prop(1), 1 / np.diff(bounds), places=14)
```
or something, in case someone changes ```Uniform``` to use some mathematically equivalent but numerically different slick new method.

Again, I'd rather have many good tests than a few perfect ones at this point, so up to you.

holopy/core/tests/test_utils_math.py
briandleahy: ....and it works!

holopy/inference/tests/test_prior.py
```diff
@@ -87,59 +347,167 @@ def test_scale_factor():
...
+
+    def test_xy_uncertainty(self):
+        expected = [Gaussian(0, 4), Gaussian(2, 4), Uniform(0, 40)]
+        evaluated = make_center_priors(self.image, xy_uncertainty_pixels = 2)
+        self.assertEqual(evaluated, expected)
+
+    def test_z_range_units(self):
+        expected = [Gaussian(0, 2), Gaussian(2, 2), Uniform(2, 10)]
+        evaluated = make_center_priors(self.image, z_range_units=(2, 10))
+        self.assertEqual(evaluated, expected)
```
briandleahy: Oops one more comment. The end of each test file should include the 2 lines:
```python
if __name__ == '__main__':
    unittest.main()
```
so they can be run from the command line individually. Nosetests is good at sniffing these out, so it doesn't matter so much, but still. We can chat if you want.

holopy/inference/tests/test_prior.py
barkls: Can we talk about this at some point? I have tried to implement but got confused...

briandleahy: Yeah! Can I swing by at 4:45 today?

briandleahy:
This looks good to me. @ralex0 did you have any feedback on this?

PR #290, opened 2019-08-16
Make tieing parameters in Scatterers more transparent

barkls:


holopy/scattering/tests/test_composite.py
```diff
@@ -0,0 +1,107 @@
+import unittest
```
briandleahy: Thanks for writing these tests. :+1:

holopy/scattering/tests/test_composite.py
```diff
@@ -0,0 +1,107 @@
...
+        n = Uniform(0, 1)
+        r = Uniform(0, np.pi/2)
+        spheres = [Sphere(n=n, r=r, center=[i, i, i]) for i in range(3)]
+        spheres = Spheres(spheres)
+        expected_keys = spheres.parameters.keys()
+        with tempfile.TemporaryDirectory() as tempdir:
+            hp.save(tempdir + '/test.out', spheres)
+            spheres = hp.load(tempdir + '/test.out')
+        self.assertEqual(spheres.parameters.keys(), expected_keys)
+
```
briandleahy: I was looking through these tests to get an idea of what ```parameters``` and ```raw_parameters``` exactly returned, and I noticed that ```raw_parameters``` isn't tested anywhere. Can we add a unit test for ```raw_parameters```?

Also, if ```raw_parameters``` is not public, it might make sense to make that a little clearer (e.g. rename it ```_raw_parameters``` if outside class don't need to use it).

holopy/scattering/scatterer/composite.py
```diff
@@ -74,45 +85,71 @@ def get_component_list(self):
...
+            parameters.update(single_scatterer_parameters)
+        return parameters
+
+    @property
+    def parameters(self):
+        parameters = self.raw_parameters
+        for tied_name, raw_names in self.ties.items():
+            for raw_name in raw_names:
+                if raw_name not in parameters.keys():
+                    msg = 'Tied parameter {} not present in raw \
```
briandleahy: When does this (and the other) error message get raised in practice? It seems that they are raised if something is not created correctly, i.e. only if there is a bug in the code here. If that's the case I think it's great to have them, but for readability you might want to make them ```assert```s rather than ```ValueError```s. If they are possible to raise by the user, then it would be nice to have some tests (at least so other developers can see what you are checking for and what fail cases are).

holopy/scattering/scatterer/composite.py
```diff
@@ -74,45 +85,71 @@ def get_component_list(self):
...
         return components
 
-    @property
-    def parameters(self):
-        pars = []
-        names = []
-        for i, scatterer in enumerate(self.scatterers):
-            for key, par in scatterer.parameters.items():
-                for index, par_check in enumerate(pars + [None]):
+    def update_ties(self):
```
briandleahy: As a general comment / question:

This seems like some pretty powerful stuff here. Is there a reason it's limited to ```Scatterers``` and not to the more general ```Scatterer``` class? For instance, I could imagine wanting to fit an ellipsoid with the two minor axes equal.

If this is not a very easy fix then let's not do it for this PR and do it for a later update when desired.

holopy/scattering/scatterer/composite.py
```diff
@@ -74,45 +85,71 @@ def get_component_list(self):
...
+                    msg = 'Tied parameter {} not present in raw \
+                           parameters {}.'.format(raw_name, parameters.keys())
+                    raise ValueError(msg)
+            tied_val = parameters[raw_names[0]]
+            for raw_name in raw_names:
+                if not parameters[raw_name] == tied_val:
+                    msg = 'Tied parameters {} and {} are not equal.'.format(
+                            parameters[raw_name], tied_val)
+                    raise ValueError(msg)
+                del parameters[raw_name]
```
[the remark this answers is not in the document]
briandleahy: OK. I'm a little confused what this function is doing. It seems that you are 
1. Copying ```raw_parameters```, which contains duplicates.
2. Removing a parameter one time when it appears in ```self.ties```.
I'm confused as to why this works, as it seems to me that this assumes that tied parameters only appear in ```raw_parameters``` two times at most (e.g. I think this will fail if there are 3 spheres all tied to share the same index). If that's the case, why doesn't something like this work?
```python
@property
def parameters(self):
    ties_reversed = {raw_name: tied_name for tied_name, raw_name in self.ties.items()}
    parameters = {
        name: parameter for name, parameter in self.raw_parameters.items()
        if name not in ties_reversed else ties_reversed[name]: parameter}
    return parameters
```

Or am I missing something?

holopy/scattering/scatterer/composite.py
barkls: It's not a trivially easy change, but it wouldn't be too hard to do. I just can't think of an obvious use case. In your example, you should just use the `spheroid` class instead of `ellipsoid`.

briandleahy: Let's just leave it as is then for now.

barkls:
I added more tests to make sure I have 100% coverage of the new code I wrote. 

I refactored a few things to remove some recursive calls and get rid of the `del` statement.

I also broke out a new method `_check_ties` since `parameters` was doing multiple different things.

Would appreciate another look from @briandleahy when you get a chance!

briandleahy:
Hey @barkls thanks for fixing these! The added tests and documentation look great!

I'm still a little confused about some of the actual code though. Can we speak tomorrow morning, say 9am-ish, about it?

barkls:
Yes

barkls:
I didn't check it compared to any other HoloPy operations, but I did
compare against an alternative way that turned out to be much slower so I
left as is.

I'm ready for merge.

Issue #254, opened 2019-03-20
data_grid raises an error when extra_dims has more than one entry


briandleahy:
Calling data_grid when ``extra_dims`` contains more than one element can sometimes raise an exception. The issues seems to arise because ``holopy.core.metadata.data_grid`` and ``holopy.core.metadata.make_coords`` assume that iteration over Python dictionaries occurs in the insertion order, which is not the case for pre-Python 3.6 versions. 

Here is a minimum working example, which raises an error on my machine:

```python
import numpy as np
from holopy.core.metadata import data_grid

ar = np.empty((1, 1, 1, 1, 2, 3, 4, 5, 6, 7, 8))
extra_dims = {
    name: np.arange(size) for name, size in zip('abcdefgh', ar.shape[3:])}

grid = data_grid(ar, spacing=1, z=0, extra_dims=extra_dims)
# ValueError: conflicting sizes for dimension 'g': length 1 on the data but length 7 on coordinate 'g'
```

Replacing the ``dict`` with an ``OrderedDict`` causes the tests to pass, 
```python
from collections import OrderedDict

import numpy as np
from holopy.core.metadata import data_grid

ar = np.empty((1, 1, 1, 1, 2, 3, 4, 5, 6, 7, 8))

extra_dims = OrderedDict(
    [[name, np.arange(size)] for name, size in zip('abcdefgh', ar.shape[3:])])

grid = data_grid(ar, spacing=1, z=0, extra_dims=extra_dims)
```
... but I'm not sure that this fixes the problem because it seems that ``holopy.core.metadata.data_grid`` and ``make_coords`` internally cast an OrderedDict to a dict anyways.

Incidentally I noticed this because some of the new tests for the 3.3 branch were failing on my machine and passing on @barkls 's.

briandleahy:
After PR #287 I ran the tests on my Python 3.5 machine and several of them failed for similar reasons on ordered vs unordered dicts. For instance, here are 2 failing tests:
```python
    bg = BoundedGaussian(*parameters.values())
  File "/home/brian/holopy/holopy/inference/prior.py", line 255, in __init__
    "Lower bound {} must be less than mean {}. Upper bound {} must"
holopy.scattering.errors.ParameterSpecificationError: Lower bound {} must be less than mean {}. Upper bound {} must be greater than mean.
```
and 
```python
    u = Uniform(*parameters.values())
  File "/home/brian/holopy/holopy/inference/prior.py", line 103, in __init__
[...]

PR #291, opened 2019-08-16
Easyissues

briandleahy:
Fixing several easy-to-fix issues in holopy.

I'm saving the bigger issue with normals and Poynting vectors for a separate pull request, as I think that will require more back-and-forth.


holopy/core/io/vis.py
```diff
@@ -261,7 +261,9 @@ def show_scatterer_slices(scatterer, spacing):
     show2d(vol)
 
 def test_display():
-    #diagnostic test to check matplotlib backend.
+    """Diagnostic test to check matplotlib backend.
+
+    You should see a white square inside a black square, with a colorbar."""
```
barkls: You should also be able to use arrow keys to switch the square into a diamond and a circle. This should be added to the description, since that is the main purpose of this function.

holopy/inference/prior.py
```diff
@@ -83,6 +84,19 @@ def unscale(self, scaled):
...
+        """
+        Uniform prior.
+
+        Parameters
+        ----------
+        lower_bouund, upper_bound : float
+        guess : float or None, optional
+            The value to take as an initial guess from the prior. If
+            guess is None, defaults to the midpoint of the prior for
+            most cases.
```
barkls: most cases means non-improper (proper?) priors. It might be good to clarify that.

holopy/inference/tests/test_prior.py
```diff
@@ -46,7 +46,7 @@ def test_construction(self):
                       'upper_bound': 3,
                       'guess': 2,
                       'name': 'a'}
-        u = Uniform(*parameters.values())
+        u = Uniform(**parameters)
```
barkls: my intention here (and on others below) was to verify the order of arguments, which is captured by * but not **. We already decided to only support python 3.7 in #254. We could also pass in matching lists of keys and values to check ordering without relying on an ordered dictionary.

holopy/inference/prior.py
```diff
@@ -211,7 +235,20 @@ def __neg__(self):
 
 class BoundedGaussian(Gaussian):
     # Note: this is not normalized
-    def __init__(self, mu, sd, lower_bound=-np.inf, upper_bound=np.inf, name=None):
+    def __init__(self, mu, sd, lower_bound=-np.inf, upper_bound=np.inf,
+                 name=None):
+        """
+        Gaussian prior restricted to an interval
```
barkls: Might be worth mentioning here that the prob and lnprob values it returns are not correctly normalized - i.e. they should be treated as relative probabilities only.

briandleahy: Should be fixed by 68438b8

holopy/inference/prior.py
briandleahy: Should be fixed by 68438b8

holopy/inference/tests/test_prior.py
briandleahy: OK. I used an OrderedDict (just for readability) and changed the test back to use ```*parameters.values()```, in commit 744dfa6.

holopy/core/io/vis.py
briandleahy: Should be fixed by 072c2e1

Issue #283, opened 2019-07-22
Holopy calculates holograms incorrectly for some/many scattering geometries


briandleahy:
The current ```calc_holo``` pipeline in ```holopy.scattering.calculations``` involves calling ```scattered_field_to_hologram```, which basically consists of this line:
```python
    holo = (np.abs(scat+ref)**2 * (1 - normals)).sum(dim=vector)
    return holo
```
This is clearly wrong -- ```1 - normals``` is not coordinate invariant. I believe what this line is trying to do is evaluate the Poynting vector, by assuming that the Poynting vector is always  E_x^2 + E_y^2, which (i) it is not (see below) and (ii) this line does not do. Two quick obvious examples why the line fails to evaluate E_x^2 + E_y^2, even when ```normals``` is normalized (which is not checked):
1. Consider the case where ```normals = (0, 0, -1)```. Then ```1 - normals = (1, 1, 2)```, and the function returns Ex^2 + Ey^2 + 2Ez^2.
2. Consider the case where ```normals = (-1, 0, 0)```. Then ```1 - normals = (2, 1, 1)```, and the function returns 2Ex^2 + Ey^2 + Ez^2. For a typical hologram, (Ex,Ey, Ez) ~ (Ex, 0, 0), so this gives a hologram which is roughly 2x what it should be.

(Incidentally, I found these behaviors by trying to generate a hologram through ```detector_points```, which currently defaults to a normal pointing in the -z direction and which gives normals approaching (-1, 0, 0) at large rho, and wondering why my hologram went to 2 at large values.)

Fortunately, this expression happens to be mostly correct (but ***still wrong***, see below) for the usual use case of a detector that is oriented along the *+z* direction. When ```normals = (0, 0, 1)```, then ```1 - normals = (1, 1, 0)```, and this returns Ex^2 + Ey^2.

One option would be to fix this buggy line by just replacing this with something like
```python
    total_field = scat + ref
    # assume the field propagates in -z
    poynting_magnitude = np.abs(total_field[:, 0])**2 + np.abs(total_field[:, 1])**2
    poynting_direction = np.array([0, 0, -1])
    poynting_flux = poynting_direction.dot(normals) * poynting_magnitude
    return poynting_flux
```

[...]

PR #293, opened 2019-08-23
Update test_calculations

ralex0:
Adding some tests as discussed in #280.  I've only added tests here but I think some of the things in calculations.py could be refactored to improve readability.

### DIFF
--- a/holopy/scattering/tests/test_calculations.py
+++ b/holopy/scattering/tests/test_calculations.py
@@ -20,6 +20,7 @@ Base class for scattering theories.  Implements python-based
 calc_intensity and calc_holo, based on subclass's calc_field
 
 .. moduleauthor:: Thomas G. Dimiduk <tdimiduk@physics.harvard.edu>
+.. moduleauthor:: R. Alexander <ralexander@g.harvard.edu>
 """
 import unittest
 
@@ -136,9 +137,17 @@ class TestPrepSchema(unittest.TestCase):
         prep_schema(LOCATIONS, MED_INDEX, wavelength, POL)
         self.assertTrue(True)
 
-class TestInterpretTheory(unittest.TestCase):
-    pass
 
+class TestInterpretTheory(unittest.TestCase):
+    def test_interpret_auto_theory(self):
+        theory = interpret_theory(SCATTERER, theory='auto')
+        theory_ok = type(theory) == Mie
+        self.assertTrue(theory_ok)
+
+    def test_interpret_specified_theory(self):
+        theory = interpret_theory(SCATTERER, theory=Mie)
+        theory_ok = type(theory) == Mie
+        self.assertTrue(theory_ok)
 
 if __name__ == '__main__':
     unittest.main()
--- a/holopy/scattering/tests/test_calculations.py
+++ b/holopy/scattering/tests/test_calculations.py
@@ -30,6 +30,7 @@ from holopy.scattering import (
 from holopy.core import detector_grid
 from holopy.core.tests.common import assert_obj_close
 from holopy.scattering.calculations import *
+from holopy.scattering.errors import MissingParameter
 
 import xarray as xr
 
@@ -37,7 +38,7 @@ SCATTERER = Sphere(n=1.6, r=.5, center=(5, 5, 5))
 MED_INDEX = 1.33
 LOCATIONS = detector_grid(shape=(20, 20), spacing=.1)
 WAVELEN = 0.66
-POL=(0, 1)
+POL = (0, 1)
 
 
 class TestCalculations(unittest.TestCase):
@@ -75,7 +76,8 @@ class TestCalculations(unittest.TestCase):
         ref = xr.DataArray(np.array([0.5, 0, 0]), coords=[('vector', coords)])
         normals = np.array((0, 0, 1))
         holo = scattered_field_to_hologram(scat, ref, normals)
-        self.assertEquals(holo.values.mean(), 1.)
+        self.assertEqual(holo.values.mean(), 1.)
+
 
 class TestDetermineDefaultTheoryFor(unittest.TestCase):
     @attr("fast")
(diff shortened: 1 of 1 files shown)


holopy/scattering/tests/test_calculations.py
```diff
@@ -30,65 +31,128 @@
...
 
-def test_calc_field():
-    field = calc_field(locations, scatterer, medium_index, wavelen, polarization)
+class TestCalculations(unittest.TestCase):
+    @attr("fast")
+    def test_calc_holo(self):
+        holo = calc_holo(LOCATIONS, SCATTERER, MED_INDEX, WAVELEN, POL)
+        self.assertTrue(True)
 
+    def test_calc_field(self):
```
barkls: would be nice to get speed attrs on all of these tests.

holopy/scattering/tests/test_calculations.py
```diff
@@ -30,65 +31,128 @@
...
     @attr("fast")
-    def test_determine_default_theory_for_on_sphere(self):
+    def test_determine_default_theory_for_sphere(self):
         default_theory = determine_default_theory_for(Sphere())
         correct_theory = Mie()
         self.assertTrue(default_theory == correct_theory)
 
     @attr('fast')
-    def test_determine_default_theory_for_on_spheres(self):
+    def test_determine_default_theory_for_spheres(self):
```
barkls: It would be nice to have a test confirming that a `Spheres` object with layered (core-shell) spheres has Mie as the default theory, since this is treated as a specific case in the `determine_default_theory_for` function.

holopy/scattering/tests/test_calculations.py
```diff
@@ -30,65 +31,128 @@
...
-def test_calc_intensity():
-    intensity = calc_intensity(locations, scatterer, medium_index, wavelen, polarization)
+    @attr("fast")
+    def test_calc_scat_matrix(self):
+        matr = calc_scat_matrix(LOCATIONS, SCATTERER, MED_INDEX, WAVELEN)
+        self.assertTrue(True)
 
+    def test_finalize(self):
+        detector = finalize(LOCATIONS.values, LOCATIONS)
+        self.assertTrue(True)
```
briandleahy: I realize that the old tests just checked that the code didn't crash, but for these tests
(test_finalize, test_calc_scat_matrix, test_calc_intensity, test_calc_cross_sections, test_calc_field, test_calc_holo) it might be good to check something more than that, given that we have the hologram. A few options are to check a particular value (maybe not great, since it might be too stringent if we change the theories), to check the return type (maybe not great if we move away from xarray) or to at least check that the return type is not None (maybe not great because it's not a very stringent test).

holopy/scattering/tests/test_calculations.py
```diff
@@ -30,65 +31,128 @@
...
 
-def test_calc_intensity():
-    intensity = calc_intensity(locations, scatterer, medium_index, wavelen, polarization)
+    @attr("fast")
+    def test_calc_scat_matrix(self):
+        matr = calc_scat_matrix(LOCATIONS, SCATTERER, MED_INDEX, WAVELEN)
+        self.assertTrue(True)
 
+    def test_finalize(self):
```
briandleahy: Can we add a ```fast```, ```medium```, or ```slow``` attr to this?

holopy/scattering/tests/test_calculations.py
```diff
@@ -30,65 +31,128 @@
...
+        args = (LOCATIONS, MED_INDEX, WAVELEN, None)
+        self.assertRaises(MissingParameter, prep_schema, *args)
+
+    def test_multiple_illumination_via_polarization_shape(self):
+        coords = ['red', 'green']
+        polarization = xr.DataArray(np.array([[1, 0], [0, 1]]),
+                                    coords=[('illumination', coords),
+                                            ('vector', ['x', 'y'])])
+        prep_schema(LOCATIONS, MED_INDEX, WAVELEN, polarization)
+        self.assertTrue(True)
```
briandleahy: Ditto with the ``assertTrue(True)``` here and below.

holopy/scattering/tests/test_calculations.py
```diff
@@ -30,65 +31,128 @@
...
-def test_calc_scat_matrix():
-    matr = calc_scat_matrix(locations, scatterer, medium_index, wavelen)
+    def test_scattered_field_to_hologram(self):
+        size = 3
+        coords = np.linspace(0, 1, size)
+        scat = xr.DataArray(np.array([0.5, 0, 0]), coords=[('vector', coords)])
+        ref = xr.DataArray(np.array([0.5, 0, 0]), coords=[('vector', coords)])
+        normals = np.array((0, 0, 1))
+        holo = scattered_field_to_hologram(scat, ref, normals)
+        self.assertEqual(holo.values.mean(), 1.)
```
briandleahy: It might be a little clearer to calculate the correct answer here, e.g.
```python
correct_holo = (np.abs(scat + ref)**2).sum(dim='vector')
self.assertEqual(holo.values[0], correct_holo.values[0])
```
or something. As is it's a little difficult to see why 1.0 is the correct answer. 

That being said, I think I have to redo this function anyway according to #283 so this might not be worth your time.

holopy/scattering/tests/test_calculations.py
ralex0: Addressed in 6aefa22 and 20c0e44

holopy/scattering/tests/test_calculations.py
ralex0: Addressed in c72bba9

holopy/scattering/tests/test_calculations.py
ralex0: Addressed in 1d6a33a and 79b6dc6.

For now we will test for specific return values.

holopy/scattering/tests/test_calculations.py
ralex0: resolved in 6aefa22

holopy/scattering/tests/test_calculations.py
ralex0: resolved by 2eb324a

holopy/scattering/tests/test_calculations.py
ralex0: resolved by 20c0e44

ralex0:
I think I was able to incorporate all of your feedback.

briandleahy:
Looks good to me. @barkls ?

holopy/scattering/tests/test_calculations.py
```diff
@@ -20,75 +20,178 @@
...
     @attr('fast')
-    def test_determine_default_theory_for_on_cylinder(self):
+    def test_determine_default_theory_for_cylinder(self):
         scatterer = Cylinder(n=1.33, h=2, d=1)
         default_theory = determine_default_theory_for(scatterer)
         correct_theory = Tmatrix()
         self.assertTrue(default_theory == correct_theory)
 
+    @attr('fast')
+    def test_determine_default_theory_for_layered_sphere(self):
```
barkls: I was actually suggesting to add a check of a `Spheres` object containing a layered sphere. Usually `Spheres` defaults to Multisphere theory, but that can't handle layered spheres, so there's a check in the `determine_default_theory_for` that looks for layered spheres and uses Mie superposition instead of multisphere if there are any present.

PR #295, opened 2019-08-24
Killnormals

briandleahy:
Deprecate normals throughout holopy.

1. This PR will cause one of the tests in #293  to fail. So let's merge #293 in first, then I can fix the 1 failing test here. But I'd like to submit this to get feedback on what I've done is good.
2. I've left ```normals``` as a valid keyword argument in the 5 user-facing functions that had it previously (```update_metadata```, ```detector_grid```, ```detector_points```, ```load_image```, and ```load_average```). However, if ```normals``` is anything other than None, a ValueError is raised with a pretty clear deprecation warning. I did this rather than completely remove normals because (1) maybe some users were using it before; if so I'd like to let them know what's up, and (2) throughout ```holopy``` things are passed in using parameter order rather than parameter names. I think I've fixed all of these, but this will give us a chance to make sure that that's the case without breaking things. Let me know if the deprecation error message is clear.

I think the way forward is for the next release, we remove the deprecation error messages and the ```normals``` keyword argument from the user facing functions.


barkls:
All looks good to me. My preference would be to only define the deprecation message once and then import it - maybe even importing a function with `if normals is not None: raise ValueError` all as one thing, but that's personal preference and not worth changing.

barkls:
The selection of only x and y components of electric field in hologram calculation is fairly messy. Try `total_field.sel(vector=['x', 'y'])` instead.

Note that this doesn't rely on having a specific dimension order or component order within the vector dimension. This is the whole point of xarray to try to make data manipulation more intuitive. Its documentation is quite good. See the table near the top of [this page] (http://xarray.pydata.org/en/stable/indexing.html) in xarray's thorough documentation for a nice summary.

PR #296, opened 2019-10-14
Issue256

### LINKED ISSUE #256: can't save Model objects with xarray properties
When saving a model by itself or as part of a results object, pack_attrs or equivalent is never run, so we have yaml.dump of xarray objects happening. These can't be loaded back up.

Possible solutions:
1) rewrite pack_attrs to work on model, etc.
2) write yaml constructor and representer classes for xarray objects

briandleahy:
Should temporarily fix #256 , by raising an error if someone tries to initialize any Model instance with an xarray.

### DIFF
--- a/holopy/core/tests/test_vis.py
+++ b/holopy/core/tests/test_vis.py
@@ -240,7 +240,9 @@ class TestDisplayImage(unittest.TestCase):
         base.attrs = {'a':2, 'b':3, 'c':4, 'd':5, '_image_scaling':(0, 59)}
         assert_equal(base.attrs, display_image(base).attrs)
 
+
 class ShowTest(unittest.TestCase):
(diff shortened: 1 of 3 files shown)


barkls:
Failing tests on travis are due to deprecation warnings from a couple sources. Not related to the contents of this PR.

briandleahy:
Do we know why the deprecation warnings are appearing now? Is it the latest version of ```emcee```? I'm a little hesitant to normalize failing tests on travis....

barkls:
Yeah, one contributer is the emcee thing so I flagged that issue with a 3.3
milestone to make sure we deal with it before release.

It also looks like h5netcdf is calling some deprecated h5py functionality,
which isn't our problem unless we have imposed some weird versioning
dependencies.

PR #297, opened 2019-10-17
Fastmielens

briandleahy:
These commits speed up the MieLens computation by another factor of 50%. Whereas before the Mielens theory was about 2x faster than the Mie theory, now MieLens is about 3x faster.

### DIFF
--- a/holopy/scattering/tests/test_mielensfunctions.py
+++ b/holopy/scattering/tests/test_mielensfunctions.py
@@ -2,6 +2,7 @@ import unittest
 import itertools
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


briandleahy:
Just a re-cap after yesterday, let's not merge this in until the Travis problems are fixed. In the mean time, I might tighten up the tests a bit since the new version is also more accurate.

barkls:
Travis passing with #298

briandleahy:
I will close and reopen this now that #298 is fixed.

ralex0:
Now you can just rebase on top of the changes @barkls made in #298 and push. Then we'll merge this PR.

PR #298, opened 2019-10-18
Make travis run without failing (and other small fixes)

barkls:


barkls:
It wouldn't be that hard for me to add a test for vis.py if you think it would help since a refactor is not imminent.

ralex0:
I'm fine with merging this now.

PR #301, opened 2019-10-30
Cleantests

briandleahy:
This PR primarily deals with some of the warnings and other stuff printed out to the console when the tests are run. Now there are only two tests which print out information to the console, ```scattering.tets.test_lowlevel.test_multisphere_singlesph``` and ```scattering.tets.test_lowlevel.test_asm```. Unfortunately, since the code that causes the print statements is in one of the C libraries ```holopy``` uses, I can't suppress this without doing something drastic. 

As an example, the test output before these commits:
```
.................S......................................................S...........................................................Exception ignored in: <_io.FileIO name=56 mode='rb+' closefd=True>
ResourceWarning: unclosed file <_io.BufferedRandom name=56>
.........................../home/brian/holopy/holopy/inference/nmpfit.py:130: UserWarning: Minimizer Convergence Failed, your results                                 may not be correct.
  may not be correct.")
....S...S..................................................................................................................................................SSSSSSSSS......... single sphere max. order:           12
 
 Estimated cluster expansion order:          14
 assembling interaction matrix row:            1
 
 Solving for incident state            1
 +iteration:            1
 error:    0.0000000000000000     
 
 Solving for incident state            2
 +iteration:            1
 error:    0.0000000000000000     
 
  Cluster expansion order:           14
... single sphere max. order:            5
 
 Estimated cluster expansion order:           9
 assembling interaction matrix row:            1
 assembling interaction matrix row:            2
 
 Solving for incident state            1
 +iteration:            1
 error:    1.6293636664628852E-002
 +iteration:            2
 error:    2.0851788369929126E-004
 +iteration:            3
 error:    4.0114965420415224E-006
 +iteration:            4
 error:    1.0340134107902293E-007
 
 Solving for incident state            2
 +iteration:            1
 error:    1.6293636664628852E-002
 +iteration:            2
 error:    2.0851788369929126E-004
 +iteration:            3
 error:    4.0114965420415224E-006
 +iteration:            4
 error:    1.0340134107902293E-007
 
  Cluster expansion order:            6
.........................................................................../home/brian/holopy/holopy/scattering/scatterer/spherecluster.py:74: OverlapWarning: Spheres(scatterers=[Sphere(n=1.59, r=5e-07, center=[3e-06, 3e-06, 1e-05]), Sphere(n=1.59, r=5e-07, center=[3.4e-06, 3e-06, 1e-05])], ties={}, warn=True) has overlaps between spheres: [(0, 1)]
  warnings.warn(OverlapWarning(self, self.overlaps))
/home/brian/holopy/holopy/scattering/scatterer/spherecluster.py:74: OverlapWarning: Spheres(scatterers=[Sphere(n=1.59, r=5e-07, center=[3e-06, 3e-06, 1e-05]), Sphere(n=1.59, r=5e-07, center=[3.4e-06, 3e-06, 1e-05])], ties={}, warn=True) has overlaps between spheres: [(0, 1)]
  warnings.warn(OverlapWarning(self, self.overlaps))
/home/brian/holopy/holopy/scattering/scatterer/spherecluster.py:74: OverlapWarning: Spheres(scatterers=[Sphere(n=1.59, r=5e-07, center=[3e-06, 3e-06, 1e-05]), Sphere(n=1.59, r=5e-07, center=[3.9e-06, 3e-06, 1e-05])], ties={}, warn=True) has overlaps between spheres: [(0, 1)]
  warnings.warn(OverlapWarning(self, self.overlaps))
/home/brian/holopy/holopy/scattering/scatterer/spherecluster.py:74: OverlapWarning: Spheres(scatterers=[Sphere(n=1.59, r=5e-07, center=[3e-06, 3e-06, 1e-05]), Sphere(n=1.59, r=5e-07, center=[3.9e-06, 3e-06, 1e-05])], ties={}, warn=True) has overlaps between spheres: [(0, 1)]
  warnings.warn(OverlapWarning(self, self.overlaps))
......................................................................S
----------------------------------------------------------------------
Ran 481 tests in 42.708s

OK (SKIP=14)
/usr/local/lib/python3.6/dist-packages/matplotlib/backends/backend_gtk3.py:197: Warning: Source ID 8 was not found when attempting to remove it
  GLib.source_remove(self._idle_draw_id)
```
And after:
```
................S......................................................S..........................................................................................S...S..................................................................................................................................................SSSSSSSSS......... single sphere max. order:           12
 
 Estimated cluster expansion order:          14
 assembling interaction matrix row:            1
 
 Solving for incident state            1
 +iteration:            1
 error:    0.0000000000000000     
 
 Solving for incident state            2
 +iteration:            1
 error:    0.0000000000000000     
 
  Cluster expansion order:           14
... single sphere max. order:            5
 
 Estimated cluster expansion order:           9
 assembling interaction matrix row:            1
 assembling interaction matrix row:            2
 
 Solving for incident state            1
 +iteration:            1
 error:    1.6293636664628852E-002
 +iteration:            2
 error:    2.0851788369929126E-004
 +iteration:            3
 error:    4.0114965420415224E-006
 +iteration:            4
 error:    1.0340134107902293E-007
 
 Solving for incident state            2
 +iteration:            1
 error:    1.6293636664628852E-002
 +iteration:            2
 error:    2.0851788369929126E-004
 +iteration:            3
 error:    4.0114965420415224E-006
 +iteration:            4
 error:    1.0340134107902293E-007
 
  Cluster expansion order:            6
...................................................................................................................................................S
----------------------------------------------------------------------
Ran 482 tests in 43.056s

OK (SKIP=14)
```

In addition, I fixed a few bugs in the tests (including one incorrectly failing test), expanded a test, and made some pep8 changes as well.

@barkls I know you're crippled right now but I'd love your feedback on this. If in person is easier lmk!


holopy/scattering/tests/test_composite.py
```diff
@@ -40,8 +40,12 @@ def test_all_ties(self):
 
     @attr("fast")
     def test_raw_parameters(self):
-        spheres = [Sphere(n=np.random.rand(), r=np.random.rand(),
-                          center=[i, i, i]) for i in range(3)]
+        max_radius = np.sqrt(3) / 2.0
```
briandleahy: @barkls just a side comment -- we were talking about whether or not to set a seed during a test; this one is a good example. Pre-3f5a79b, the max radius was (implicitly) 1 which is > sqrt(3) / 2 and could lead to some overlaps. However, it just happened that the radii that were drawn were less than sqrt(3) /2, so the old version of this test would pass. It would pass every time because the tests are run in the same order and the random seed is set by some of the tests. However, when I factored out a different test, it changed the order the tests were run in, which changed the random values of the radii and happened to lead to a failure. Cool story, huh?

holopy/inference/tests/test_minimizer.py
```diff
@@ -74,12 +88,31 @@ def test_iter_limit():
...
+                prior.Uniform(5, 25, 20)),
+        r=(prior.Uniform(.4, .6, guess = .45)),
+        n=1.59)
+    par_s = Spheres([guess1, guess2])
 
-    model = AlphaModel(par_s, medium_index=1.33, illum_wavelen=.66, illum_polarization=(1, 0), alpha = prior.Uniform(.1, 1, .6))
+    model = AlphaModel(
+        par_s, medium_index=1.33, illum_wavelen=.66, illum_polarization=(1, 0),
+        alpha=prior.Uniform(.1, 1, .6))
     warnings.simplefilter("always")
```
barkls: do you know why this is here? It seems an odd spot to be messing around with warnings. When running tests, HoloPy is configured to have warnings raise exceptions. It's probably not a good idea to have individual tests alter this behaviour unless using `with warnings.catch_warnings*()` as you do immediately below.

Can you delete it without making tests fail?

holopy/scattering/tests/test_composite.py
barkls: Neat find! Thanks for fixing.

holopy/inference/tests/test_minimizer.py
briandleahy: Should be fixed by af57daf.

However there are 2 other spots where a ```simplefilter``` is set w/o a catch statement:

1. In ```holopy.core.tests.common```, it says:
```python
# tests should fail if they give warnings
import warnings
warnings.simplefilter("error")
```
2. In ```holopy.scattering.scatterer.spherecluster``` (i.e. not in a test), it says:
```python
# default to always warning the user about overlaps.  This can be overriden by
# calling this function again with a different action.
warnings.simplefilter('always', OverlapWarning)
```

I think the second is correct, but I'm not sure about the first. Is that what we want? How do we know that ```core.tests.common``` is imported first?

barkls: You raise a good point. I think the intention was that all test files import something from `holopy.core.tests.common`, thereby running that line of code. There are currently 9 test files in HoloPy that do not follow that pattern, most of which we recently created/refactored:

```
holopy/core/tests/test_errors.py
holopy/inference/tests/test_cma.py
holopy/inference/tests/test_prior.py
holopy/scattering/tests/test_composite.py
holopy/scattering/tests/test_lowlevel.py
holopy/scattering/tests/test_mielens.py
holopy/scattering/tests/test_mielensfunctions.py
holopy/scattering/tests/test_scatterer.py
holopy/scattering/tests/test_tmatrix.py
```
I'm not sure what the best solution is...

barkls:
Hey @briandleahy I have some code already in holopy (probably in multisphere?) that suppresses the print statements. It might be worthwhile to factor this out as a context manager and use that to suppress the additional printing during tests.

barkls:
it's in ```holopy.scattering.theory.multisphere```, lines 186-206

holopy/inference/tests/test_minimizer.py
briandleahy: Will be solved by #302

briandleahy:
Should be fixed. Now the test output is much tidier:

```
nosetests3 -a "fast"
.....................................................................S................................................................................S...S....................................................................................................................S......................................................................................................................
----------------------------------------------------------------------
Ran 390 tests in 2.277s

OK (SKIP=4)
```

I have a 2 questions / comments in the code before merging; see above.

holopy/core/utils.py
```diff
@@ -39,6 +39,29 @@
 
 from holopy.core.errors import DependencyMissing
 
+
+# FIXME this is difficult to test, since it works on the file level
```
briandleahy: @barkls  and @ralex0 -- I don't know how to test this effectively. There are ways of testing a normal print statement (e.g. contextlib.redirect_stdout), but I can't find one that works for output generated by external code. Do either of you have any thoughts?

holopy/scattering/theory/multisphere.py
```diff
@@ -183,27 +184,15 @@ def _scsmfo_setup(self, scatterer, medium_wavevec, medium_index):
...
-            -1.0 * centers[:,2],  m.real, m.imag,
-            scatterer.r * medium_wavevec, self.niter, self.eps,
-            self.qeps1, self.qeps2,  self.meth, (0,0))
-
-        if self.suppress_fortran_output:
-            #restore stdout to default and close files
-            os.dup2(default,1)
-            os.close(devnull)
-            os.close(default)
+        with SuppressOutput(suppress_output=self.suppress_fortran_output):
```
briandleahy: This line is pretty ugly, because it needs the output to not suppress fortran output. Does anyone actually want to see the fortran output when they are calling scattering theory? Can I just make the context manager suppress output and not have the misdirection?

holopy/core/utils.py
vnmanoharan: Is the problem that the fortran code is outputting too much information?  If so, we should check to see if the code is outputting to stdout or stderr.  By UNIX convention, diagnostic information should be printed to stderr rather than stdout.  Then you can filter the correct stream (or change the fortran code to output to stderr).

briandleahy: The problem is just that the fortran code outputs extraneous info, which makes it more difficult to see which tests have passed and failed (among other things). It's just a fortran ```print*``` statement (line 146ff of ```holopy.scattering.theory.mie_f.scsmfo_min.for```) so I assume it's outputting to stdout. I don't think stderr is the place for it, since it's not really error related, it's just ... noise.

I'd be happy to edit the fortran code and delete the offending lines, but I wasn't sure if other holopy users use them, so I figured I'd just use that context manager to pipe the output to devnull.

Do you think it's better to delete the print statements from the fortran code, or leave them there and capture it?

barkls: I agree stderr isn't appropriate. I have never made use of the terminal output from the multisphere code when I run it. In fact, I had those lines commented out of my local fortran file for a long time until I wrote the current code that suppresses it, so I'd definitely be on board to delete them.

vnmanoharan: stderr is the appropriate stream.  stdout is not.  Any output from the program that doesn't go directly into the stdin for another program should be directed to stderr.  That includes diagnostic information.  We should not delete the lines from the fortran code that output this information.  It's not noise -- it gives information about the convergence of the conjugate gradient solver.  But you can direct it to stderr and then redirect to /dev/null if you don't want it.

barkls:
In conversation with @briandleahy, we agreed to keep the Fortran output in stdout vs. stderr, since it's usually suppressed anyway. The context manager gives a convenient way to to hide/expose the output as necessary.

briandleahy:
Thanks @barkls . I think maybe down the line we can edit the fortran code to output to stderr, but I don't think it's urgent now.

PR #302, opened 2019-11-01
add nose plugin to treat warnings as exceptions

barkls:
There are a few changes here:

1) A nose plugin is defined in `holopy.core.tests.common.py` that raises exceptions when warnings are encountered when running tests.

2) When `setup.py` is run, this plugin is also installed globally (unfortunately this is how nose handles plugins; nose2 has fixed this). This will work if `setup.py` is run with either `install` or `develop` 
commands. 
**edit**: it will not work when running `python setup.py build_ext --inplace` as we previously recommended in our documentation. see #303.

3) The new `setup.cfg` file runs nosetests with the plugin activated whenever nosetests is run from the holopy root directory. Unfortunately it won't activate automatically if run from subdirectories.

4) nose is now imported in `setup.py`, and so it has become a build dependency and not just a test dependency.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,6 @@ from numpy.distutils.core import setup, Extension
 
 HOLOPY_NOSE_PLUGIN_LOCATION = ('holopycatchwarnings = '
                                'holopy.core.tests.common:HoloPyCatchWarnings')
-print(HOLOPY_NOSE_PLUGIN_LOCATION)
 
 hp_root = os.path.dirname(os.path.realpath(sys.argv[0]))
 
--- a/setup.py
+++ b/setup.py
@@ -37,8 +37,9 @@ from os.path import join
 import nose
 from numpy.distutils.core import setup, Extension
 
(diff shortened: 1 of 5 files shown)


barkls:
I have no idea how to test this. Let me know if you think of anything!

setup.py
```diff
@@ -33,8 +33,13 @@
 import os
 import sys
 from os.path import join
+
+import nose
 from numpy.distutils.core import setup, Extension
 
+HOLOPY_NOSE_PLUGIN_LOCATION = ('holopycatchwarnings = '
+                               'holopy.core.tests.common:HoloPyCatchWarnings')
+print(HOLOPY_NOSE_PLUGIN_LOCATION)
```
briandleahy: Why does ```setup.py``` print ```HOLOPY_NOSE_PLUGIN_LOCATION```?

barkls: Good catch. Removed.

PR #304, opened 2019-11-11
Postinstall for win64

ralex0:
To get our win64 build working with conda-forge, we need to make the setup.py compatible with Microsoft Visual Studio (MSVC) 2015 (14.0)'s C compiler. I have learnt that for some [magical] reason when the C compiler is MSVC2015, when f2py runs, it puts the dynamic link libraries (dll) required for holopy's fortran extensions to run in holopy/.lib. [On a whim](https://stackoverflow.com/questions/48826283/compile-fortran-module-with-f2py-and-python-3-6-on-windows-10) I decided to put these libraries in their "appropriate" locations next to their `.pyd` files within the package. This allowed me to import the extensions compiled with MSVC2015 as the C compiler and msys2's gfortran as the Fortran compiler. [Here is the conda build environment in which I discovered this](https://github.com/manoharan-lab/holopy/files/3829356/holopy-win.zip). I have [mingw64 installed through chocolatey](https://chocolatey.org/packages/msys2), so that's why there's no `m2w64-gcc-fortran` in the environment.

This PR moves these libraries to the appropriate locations in a post install script if the platform is Windows.

Theoretically there is a different workaround using the `f2py_options` keyword argument of `numpy.distutils.core.setup`, but I couldn't make this work. The option that I want to pass is `--compiler=mingw32`. On my win64 environments that match one conda-forge uses (i.e. using MSVC2015) I can't build **any** f2py extension unless I either pass `--compiler=mingw32` or just move the dlls like in this patch. The problem is passing the `f2py_options` in setup.py does nothing as far as I can see.

Since these changes are changes to the way holopy is built, I'm not sure how to test it.


barkls:
It's great you got this to build! It's too bad the f2py_options method doesn't work, since that seems 
so much more straightforward and less likely to break if HoloPy is installed to an unexpected location.

Where did you try adding the  `f2py_options` flag? It looks like the `config.add_extension` calls would be the most basic place, but somewhere higher up might also work?

What format did you use for the `f2py_options` argument? It looks like it can be tricky to get it to parse correctly, but it should possibly be a list of strings: `f2py_options=['compiler:', 'mingw32', ':']`.

I might play around with this later on my Windows laptop, but I want to make sure I'm not redoing things you already tried!

ralex0:
I tried passing as a dictionary to `config.add_extension`. I'll experiment as well tonight because I agree that passing `--compiler=mingw32` to `f2py_options` is the better solution. I think you're onto something with that syntax which I did not try yet.

ralex0:
We might want to include this patch even if we get `f2py_options` to work as intended. Maybe we should change this PR to check if MSVC is installed instead of whether `os.name == 'nt'`. 

holopy can cetrainly be compiled without MSVC, but currently conda-forge insists that MSVC is the C compiler that must be used on Windows in their DevOps pipeline. [IMO] The end-user shouldn't have to worry about their environment, so if they just so happen to have MSVC installed, holopy will still work for them with this patch.

This post-install script basically does nothing even if the OS in Windows if the dlls don't exist. Furthermore, any Windows user that is able to compile holopy right now _probably_ has gfortran, so it's necessary that they'll also have g++, and the "--compiler=mingw32" keyword will succeed if we can figure it out. However, an edge case is the user has MSVS14 as the C compiler and ifort (Intel) as the Fortran compiler.

Another thing is we would definitely need to keep `m2w64-toolchain` as a host requirement for win in `meta.yaml` since there is no way to ensure the host has gcc  unless we have more control over provisioning, which is not possible with conda-forge's build pipeline as far as I know. Ideally, I would run `choco install msys2` before the install. 

Thoughts?

barkls:
I tried to play around with this on my computer, but I'm afraid it is beyond me.  Did you get a chance to try the `f2py_options` passed in as a list of strings?

Even if we can't get that working, this PR lets conda-forge build HoloPy, which is all we need for a distribution. It should also be fairly straightforward for developers to install as well since MSVC is readily available. The developer documentation should be updated to reflect this change. Do you feel like doing that as part of this PR?

I do think it's a good idea to check if MSVC is installed rather than checking operating system.

If we can get a gcc-based build working, I prefer keeping the `m2w64-toolchain` as a requirement in meta.yaml vs. running installations through other package managers - that seems very opposed to the goals of conda-forge and is likely to run into issues as they update things.

setup.py
```diff
@@ -29,22 +29,40 @@
...
 hp_root = os.path.dirname(os.path.realpath(sys.argv[0]))
 
-# this will automatically build the scattering extensions, using the
-# setup.py files located in their subdirectories
-def configuration(parent_package='',top_path=' '):
+class PostDevelopConfig(develop):
+    """Post-installation for development mode."""
+    def run(self):
+        develop.run(self)
+        if os.name == 'nt': _move_S_msvc_libs('develop')
```
barkls: This function doesn't exist. I think it should just be `_move_msvc_libs`

setup.py
```diff
@@ -29,22 +29,40 @@
...
+
+
+class PostInstallConfig(install):
+    """Post-installation for installation mode."""
+    def run(self):
+        install.run(self)
+        if os.name == 'nt': _move_msvc_libs('install')
```
barkls: Is there a way to combine these functions into one? They have a lot of repeated code...

setup.py
```diff
@@ -62,10 +80,36 @@ def configuration(parent_package='',top_path=' '):
...
+    tmatrix_f_dir = os.path.join(package_dir, 'scattering', 'theory', 'tmatrix_f')
+    mie_f_dir = os.path.join(package_dir, 'scattering', 'theory', 'mie_f')
+
+    for dll in tmatrix_libs: shutil.move(dll, tmatrix_f_dir)
+    for dll in mie_libs: shutil.move(dll, mie_f_dir)
+    shutil.rmtree(lib_dir)
+
+def _get_holopy_install_dir(mode):
+    if mode == 'install':
+        sitepackages = list(site.getsitepackages())
```
barkls: How robust is this way of finding where the package is installed?

ralex0: Well, it works on CI, so I stay a step above "works on my machine." 

For install mode AFIAK the only other directory that `site.getsitepackages()` returns is where the python executable is installed. 

For develop mode, there might be some case I'm missing.

setup.py
ralex0: I took a look at the source of setuptools, and the `install` and `develop` classes really are quite different (not even the same base class) even though they have the same API. Maybe there's something clever that we can do like have some class devrive`develop` and `install`

setup.py
ralex0: Yeah, I updated this in @5fdef54

setup.py
ralex0: Refer to the official documentation for [python 3.6](https://docs.python.org/3.6/using/windows.html#finding-modules) and [python 3.7](https://docs.python.org/3.7/using/windows.html#finding-modules). On Windows, `site.getsitepackages()` returns two elements: the default python library, and the third-party library. 

According to the docs, the third party library is a sub-directory of the default library called "site-packages". The next line `path for path in site.getsitepackages() if 'site-packages' in path` selects this third party directory for the install location of holopy __if and only if the user has not manually edited their python path__.  Actually, now that I am re-reading it, line 105 is unneeded.

Therefore, unless the user has modified their python installation manually, this routine works on Windows. Also the routine only executes on Windows. I don't know why your site-package list includes libraries that are missing,.

ralex0: Also, I have no interest in trying to support *any* arbitrary python installation. If the user is that picky about how python is installed, they probably aren't on Windows. Even if they are, they probably know how to fix this. Otherwise, they should raise an issue and I'll address it.

setup.py
```diff
@@ -62,10 +80,36 @@ def configuration(parent_package='',top_path=' '):
...
+        dir = [path for path in site.getsitepackages() if 'site-packages' in path]
```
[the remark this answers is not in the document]
ralex0: See @c76ce7c

setup.py
```diff
@@ -62,10 +80,36 @@ def configuration(parent_package='',top_path=' '):
...
-    __version__ = 'unknown'
+def _move_msvc_libs(mode='install'):
+    """ These dlls need to be moved if the fortran is complied in an environment
+    with MSVC 2015 as the C compiler.
+    """
+    package_dir = _get_holopy_install_dir(mode)
+    lib_dir = os.path.join(package_dir, '.libs')
+    sep = os.path.sep
+
+    tmatrix_libs = glob.glob(lib_dir + sep + 'libS.*.dll')
```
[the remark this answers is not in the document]
ralex0: See c76ce7c

setup.py
```diff
@@ -29,22 +29,40 @@
...
+    """Post-installation for installation mode."""
+    def run(self):
+        install.run(self)
+        if os.name == 'nt': _move_msvc_libs('install')
+
+
+def configuration(parent_package='',top_path=''):
+    # this will automatically build the scattering extensions, using the
+    # setup.py files located in their subdirectories
     from numpy.distutils.misc_util import Configuration
```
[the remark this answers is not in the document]
ralex0: See c76ce7c

setup.py
```diff
@@ -29,22 +29,40 @@
 "gfortran" and "python-dev" packages installed.
 '''
 
-import setuptools
-import os
-import sys
+import glob, os, setuptools, shutil, site, sys
```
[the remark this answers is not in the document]
ralex0: See c76ce7c

ralex0:
245fd17 is based on a conversation @barkls and I had yesterday

setup.py
briandleahy: OK, sounds good @ralex0 . Can you just put a line or so of this discussion in a comment near the code, so if a problem does arise in the future there will be an easier time debugging this?

Otherwise I'm happy if you think this will work for every "generic Python user".

briandleahy:
As long as this works on your Windows machine from a fresh install, this looks good to me @ralex0 ! Thanks for doing this. 

I'll let @barkls merge this in, in case he has any other questions or comments.

setup.py
ralex0: Yeah this works in my fresh windows environments.

Issue #230, opened 2018-06-07
no wrapper for xarray.DataArray


barkls:
When we deleted Marray, we switched to using xr.DataArray objects to represent holograms. We chose to use these objects as is, without defining a wrapper class.

Now, it seems like we might want a wrapper class for a few reasons:
1) There are many small functions that act on holograms that might be better as methods. Currently, these are scattered between metadata.py, io.py, vis.py, and others.
2) This would allow us to standardize the format of hologram objects, so we wouldn't need to do as many checks to see if a given hologram has a particular attribute or has values in a specific format.
3) Printing or inspecting DataArrays and their components is always really messy, so we could write some nice functions to help there.
4) detector_grid and similar functions are unintuitive because they seem like they should be instantiations of a class. We could convert these functions to _init_ functions of a wrapper class

Note that this change would require some rewriting and changes to API, so it would have to be a major release.

korigod:
I think it's possible to introduce some changes without breaking the API if the wrapper is inherited from `xArray.DataArray`. It's possible to add some convenient methods, override `__str__` and mark the old API as deprecated but leave it until v4. Or do you think about composition instead of inheritance here?

barkls:
I think you're right, but the truth is I haven't given enough thought to what our goals with this are. I think we would want to remove our dependence on the DataArray.attrs attribute and just define our own attributes for things like illum_wavelen, noise_sd, etc. on a hologram. Many of these changes would happen in the background, so they probably wouldn't affect the API. However, I don't really want to proceed on this until I have a better sense of what we are trying to accomplish.

briandleahy:
My 2 cents:

To me, the bigger issue is that the methods for creating and modifying ```xarray```s are scattered and work independently of each other (i.e. all the code in metadata). Since the code is not the same for these different creational pathways, it's possible to get ```xarray```s with different properties when they are created from the same pipeline. This makes dealing with different ```xarray```s trickier than it could be. For instance, ```detector_points``` and ```make_subset_data``` both create ```xarray```s which are an unstructured set of data, but those two functions don't use the same sub-methods. Likewise, ```detector_grid``` doesn't enforce what the ```extra_dims``` are; they can be things that should be there like wavelength, or they can be irrelevant things holopy doesn't support, or there can be things which are used keys but formatted incorrectly. This multiplicity of functionality leads a lot of code checking whether an ```xarray``` has the right attributes, and it leads to to little bugs proliferating throughout the code -- things like issue #289.

[...]

PR #305, opened 2019-11-14
Update inference api

### LINKED ISSUE #265: MCMC is not the same as fitting
In 3.3, both MCMC and least-squares fitting have been rolled into `Strategy` objects, and the main method for MCMC has been renamed from `sample()` to `optimize()`.  But MCMC is not an optimization method, and the type of information returned from running the sampler is not the same as the type of information returned from running an optimization algorithm.  Optimization and sampling serve different purposes: optimization returns a point estimate and sampling can be used to estimate the entire posterior distribution.  While both are "strategies" in some sense, they are strategies to achieve two different ends.  We need to rethink how fitting should enter into the object hierarchy.

barkls:
Fixes #265

[Here are the docs](https://holopy.readthedocs.io/en/update-inference-api/) of this branch.


docs/source/users/fit_tutorial.rst
```diff
@@ -38,121 +36,363 @@ We start by loading and processing data just as we did for the parameter inferen
...
+       center.2 = 1.089919325  
+    Iter      15    CHI-SQUARE =  410303.9207  DOF =  39997
+       r = 0.9559566861  
+       center.0 = 1.006806381  
+       center.1 = 0.9928656724  
+       center.2 = 1.089919691  
+
+Here we have used an :class:`.ExactModel` which takes a function ``calc_func``
+to apply on the :class:`.Scatterer` (we have used :func:`.calc_holo` here).
+The :class:`.ExactModel` is the default for :func:`.fit` but HoloPy also
```
briandleahy: My 2 cents is that the `fit` function should have defaults that are the most generally helpful. Since we currently need an alpha to fit holograms, I think that `fit` should default to `AlphaModel`, not `ExactModel`.

holopy/core/utils.py
```diff
@@ -161,6 +162,25 @@ def repeat_sing_dims(indict, keys = 'all'):
 
     return updated(indict, subdict)
 
+class LnpostWrapper(HoloPyObject):
```
briandleahy: Is this tested in holopy right now?

I realize that you're just moving code from one file to another, so writing unit tests for this is above and beyond the call of duty. Do we want to make a list of objects though that aren't tested, or no?

holopy/inference/cmaes.py
```diff
@@ -68,10 +67,12 @@ class CmaStrategy(HoloPyObject):
         number of threads to use or pool object or one of {None, 'all', 'mpi'}.
         Default tries 'mpi' then 'all'.
     """
-    def __init__(self, npixels=None, resample_pixels=True,
+    def __init__(self, npixels=None, popsize = None, resample_pixels=True,
```
briandleahy: pep8 :smiley:

holopy/inference/emcee.py
```diff
@@ -46,26 +45,38 @@ def sample_one_sigma_gaussian(result):
 
 
 class EmceeStrategy(HoloPyObject):
-    def __init__(self, nwalkers=100, npixels=None, parallel='auto', cleanup_threads=True, seed=None, resample_pixels=False):
+    def __init__(self, nwalkers=100, nsamples=1000, npixels=None, walker_initial_pos = None, parallel='auto', cleanup_threads=True, seed=None):
```
briandleahy: pep8 again :smiley:

holopy/inference/cmaes.py
```diff
@@ -80,25 +81,29 @@ def __init__(self, npixels=None, resample_pixels=True,
...
         if self.npixels is not None and self.new_pixels is None:
             data = make_subset_data(data, pixels=self.npixels, seed=self.seed)
-        if popsize is None:
-            numpars = len(parameters)
-            popsize = int(2 + numpars + np.sqrt(numpars)) #cma uses 4+3*ln(n)
-        if walker_initial_pos is None:
-            walker_initial_pos = model.generate_guess(popsize, seed=self.seed)
+        if self.popsize is None:
+            npars = len(parameters)
+            self.popsize = int(2 + npars + np.sqrt(numpars))#cma uses 4+3*ln(n)
```
briandleahy: This is a typo, I believe. `npars` is defined but now `numpars` is not.

If this is the case, can we capture this failure with a unit test before fixing it?

holopy/inference/functional_api.py
```diff
@@ -0,0 +1,76 @@
+# Copyright 2011-2016, Vinothan N. Manoharan, Thomas G. Dimiduk,
```
briandleahy: The code in this file looks great, but can we run it through a pep8 checker? I use pep8online.com

One other regarding pep8 -- per pep8, module names should not have an underscore. Can we rename this from ```functional_api``` to something without an underscore?

holopy/inference/result.py
```diff
@@ -135,6 +120,24 @@ def add_attr(self, kwargs):
             setattr(self, key, val)
             self._kwargs_keys.append(key)
 
+    def forward(self, pars):
+        if hasattr(self.data, 'original_dims'):
```
briandleahy: Oof! This is a reason why, down the line, we should address #230 and ensure that `detector_grid`, `detector_points`, and `subset_data` all play the same way. Not really a comment on this PR though :smiley:

holopy/inference/tests/common.py
```diff
@@ -0,0 +1,39 @@
...
+# You should have received a copy of the GNU General Public License
+# along with HoloPy.  If not, see <http://www.gnu.org/licenses/>.
+
+import numpy as np
+
+from holopy.inference import prior
+from holopy.inference.model import Model
+from holopy.scattering import Sphere
+
+class SimpleModel(Model):
```
briandleahy: Why is SimpleModel defined both here and in `test_cma`? Am I missing something?

holopy/inference/tests/test_functional_api.py
```diff
@@ -0,0 +1,137 @@
...
+        with warnings.catch_warnings():
+            warnings.simplefilter('ignore', UserWarning)
+            result = fit(DATA, model, ['r', 'y'])
+        self.assertEqual(result._names, ['n', 'center.0'])
+
+    @attr('medium')
+    def test_passing_model_and_parameters_gives_warning(self):
+        model = make_default_model(SPHERE, None)
+        self.assertWarns(UserWarning, fit, DATA, model, ['r', 'y'])
+
```
briandleahy: Can we get a `def test_sample_raises_error_when_invalid_model`? Otherwise I think you hit all the directions of the functional api.

holopy/inference/tests/test_model.py
```diff
@@ -84,17 +88,72 @@ def test_yaml_round_trip_with_dict(self):
...
                 dims=['illumination'],
                 coords={'illumination': ['red', 'green']})
             kwargs = {key: value}
-            model = BaseModel(sphere, **kwargs)
+            model = Model(sphere, **kwargs)
             with self.subTest(key=key):
                 reloaded = take_yaml_round_trip(model)
                 self.assertEqual(reloaded, model)
 
+class TestModelFittingMethods(unittest.TestCase):
```
briandleahy: :+1:

holopy/core/utils.py
ralex0: Ditto what @briandleahy said

holopy/inference/functional_api.py
ralex0: I use [pycodestyle](https://github.com/PyCQA/pycodestyle) to check pep8 before committing.

docs/source/users/fit_tutorial.rst
barkls: The whole point is that the functional interface it's easy to use, so I made the change you suggested.

holopy/core/utils.py
barkls: I confirmed there are no tests for this object, but I'm not writing them for this PR.

holopy/inference/cmaes.py
barkls: done.

holopy/inference/emcee.py
barkls: done.

holopy/inference/cmaes.py
barkls: done. There are embarrasingly few tests for the cma.py module, but I added this one!

holopy/inference/functional_api.py
barkls: done. I renamed it to "inference/interface.py" and also renamed scattering/calculations.py to "scattering/interface.py" for consistency. Let me know if you have any concerns about this!

holopy/inference/result.py
barkls: This isn't as bad as you think! It's checking if the result was calculated only on a subset of pixels, but to display the hologram you want to calculate it all the pixels in the original image, even if the results object doesn't know about them.

I don't think we can remove the if-check here without assigning some properties to the objects storing holograms, which would mean not using raw xarray.DataArrays. I agree that's a valid long-term goal but this isn't as straightforward as streamlining the functions to generate them.

holopy/inference/tests/common.py
barkls: I moved it but forgot to remove the original. Fixed.

holopy/inference/tests/test_functional_api.py
barkls: That is actually the first test in `test_interface`:

```python
    @attr('fast')
    def test_cannot_sample_without_model(self):
        self.assertRaises(ValueError, sample, DATA, Sphere())
```
unless I am misunderstanding the test you are requesting...

You did inspire me to add a test that the `fit` function doesn't work with arbitrary input.

barkls:
Thanks for the review @briandleahy and @ralex0! I addressed all your changes - let me know if you any of them weren't what you had in mind or if you have any other requests.

@vnmanoharan are you satisfied with this?

holopy/scattering/theory/tmatrix.py
```diff
@@ -29,7 +29,11 @@
 from holopy.scattering.errors import TheoryNotCompatibleError, TmatrixFailure
 from holopy.core.errors import DependencyMissing
 from holopy.scattering.theory.scatteringtheory import ScatteringTheory
-from holopy.scattering.theory.tmatrix_f.S import ampld
+try:
+    from holopy.scattering.theory.tmatrix_f.S import ampld
+    COMPILED_TMATRIX_FORTRAN = True
+except:
```
vnmanoharan: Make this specific to the `ModuleNotFoundError` exception.  Although right now the `import` statement has only one exception, it might change in the future, and it's good practice to catch only specific exceptions when you're passing control back to the program after catching.

holopy/inference/tests/test_interface.py
```diff
@@ -0,0 +1,144 @@
+# Copyright 2011-2016, Vinothan N. Manoharan, Thomas G. Dimiduk,
+# Rebecca W. Perry, Jerome Fung, and Ryan McGorty, Anna Wang
```
vnmanoharan: Add your names to the copyright

holopy/fitting.py
```diff
@@ -78,32 +77,32 @@ def __new__(self,guess=None, limit=None, name=None, **kwargs):
...
 
-class Parametrization(HoloPyObject):
+class Parametrization():
     def __new__(self, scatclass, pars):
         fit_warning('Scatterer', 'Parametrization')
         return ParameterizedObject(scatclass(*pars))
 
 
-class limit_overlaps(HoloPyObject):
+class limit_overlaps():
```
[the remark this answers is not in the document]
barkls: This entire file exists only to preserve backwards compatibility and will be deleted in an upcoming version of HoloPy. We need to keep the namespace exactly as it was, but I could make this a function instead of a class, since it behaves like one anyway. I ran into problems doing that with some other classes because they couldn't be used to instantiate objects from saved data values if they were functions. I don't know whether that specific issue would be a problem for `limit_overlaps` but it could be.

To further complicate things, `limit_overlaps` is a function (not a class) on the current master branch. It was changed to a class very early in the 3.3 development cycle for compatibility with parallel computation. The original name was maintained then against PEP8 to avoid breaking things because it was anticipated as a minor 3.2.2 release, and not as 3.3. So any potential problems with loading HoloPy objects that contain `limit_overlap` objects would be limited to users of the pre-release 3.3 branch.

I think it's safest to keep the version currently in the PR as the file will soon be deleted anyway, but there doesn't seem to be a big risk in making this a function if you prefer.

docs/source/users/fit_tutorial.rst
```diff
@@ -38,121 +36,368 @@ We start by loading and processing data just as we did for the parameter inferen
...
-  hp.save('result.h5', result)
 
-.. _random_subset:
+Customizing the model
+~~~~~~~~~~~~~~~~~~~~~
+Sometimes you might want a bit more control over how the parameters are varied.
+You can customize the parameters with a :class:`.Model` object that describes
+parameters as :class:`.Prior` objects instead of simply passing in your best
+guess scatterer and the names of the parameters you wish to vary. For example,
+we can set bounds on the coordinate parameters and and use a Gaussian prior
```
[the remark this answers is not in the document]
barkls: fixed

docs/source/users/fit_tutorial.rst
```diff
@@ -38,121 +36,368 @@ We start by loading and processing data just as we did for the parameter inferen
...
+(std=0.05) for radius.
```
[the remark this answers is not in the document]
barkls: fixed.

docs/source/users/fit_tutorial.rst
```diff
@@ -38,121 +36,368 @@ We start by loading and processing data just as we did for the parameter inferen
...
-``imag = par(1e-4)`` instead of ``imag = 1e-4``. In a case like this
-where we are providing a small imaginary part for numerical stability,
-you would not want to fit to it. However fitting to an imaginary index
-component could be useful for a metal particle. Setting the key word argument ``step = 0.01`` specifies the the step size used in calculating
-the numerical derivatives of this parameter. Specifying a small step
-size is often necessary when fitting for an index of refraction.
-
-Tying Parameters
-----------------
+When this is used to define a :class:`.Sphere`, :func:`.fit` will fit to
```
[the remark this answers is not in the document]
barkls: fixed.

docs/source/users/fit_tutorial.rst
```diff
@@ -38,121 +36,368 @@ We start by loading and processing data just as we did for the parameter inferen
...
-  sc = Spheres([Sphere(n = n1, r = par(0.5e-6), \
-    center = [10., 10., 20.]), \
-    Sphere(n = n1, r = par(0.5e-6), center = [9., 11., 21.])])
+    n1 = prior.Gaussian(1.58, 0.02)
+    sphere_cluster = Spheres([
+    Sphere(n = n1, r = 0.5, center = [10., 10., 20.]),
+    Sphere(n = n1, r = 0.5, center = [9., 11., 21.])])
+
+
+Sampling a Posterior
```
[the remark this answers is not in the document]
barkls: renamed to "Bayesian Parameter Estimation"

docs/source/users/fit_tutorial.rst
```diff
@@ -38,121 +36,368 @@ We start by loading and processing data just as we did for the parameter inferen
...
+Customizing the algorithm
+~~~~~~~~~~~~~~~~~~~~~~~~~
+The :func:`.fit` and :func:`.sample` functions follow algorithms that determine
+which sets of parameter values to simulate and compare to the experimental
+data. You can specify a different algorithm by passing a *strategy* keyword
+into either function. Options for :func:`.fit` currently include the default
+Levenberg-Marquardt (``strategy="nmpfit"``), as well as cma-es
+(``strategy="cma"``) and scipy least squares (``strategy="scipy lsq"``).
+Options for :func:`.sample` include the default without tempering
+(``strategy="emcee"``) or tempering by changing the number of pixels evaluated (``strategy="subset tempering"``) or Monte Carlo temperature
```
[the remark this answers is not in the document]
barkls: fixed.

docs/source/users/fit_tutorial.rst
```diff
@@ -38,121 +36,368 @@ We start by loading and processing data just as we did for the parameter inferen
...
+---------------------
+In the most recent example, we evaluated the holograms at the locations of only
+500 pixels in the experimental image. This is because a hologram usually
+contains far more information than is needed to estimate your parameters of
+interest. You can often get a significantly faster fit with little or no loss
+in accuracy by fitting to only a random fraction of the pixels in a hologram. 
+
+You will want to do some testing to make sure that you still get
+acceptable answers with your data, but our investigations have shown
+that you can frequently use random fractions of .1 or .01 with little
```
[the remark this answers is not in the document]
barkls: fixed.

docs/source/users/load_tutorial.rst
```diff
@@ -126,7 +126,7 @@ specify metadata when loading the image, we can use :func:`.update_metadata` to
     holo = hp.core.update_metadata(holo, medium_index=1.33, illum_wavelen=0.660, illum_polarization=(1,0))
 
 .. note::
-    Spacing and wavelength must both be written in the same units - microns in the example
+    Spacing and wavelength must be given in the same units - microns in the example
```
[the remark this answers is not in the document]
barkls: fixed.

holopy/scattering/theory/tmatrix.py
barkls: fixed.

holopy/inference/tests/test_interface.py
barkls: I added my name to all files I have contributed to.

holopy/fitting.py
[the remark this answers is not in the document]
vnmanoharan: OK to keep as is.  Mark with a brief comment covering what you wrote above.  Rest is ready to merge.

PR #306, opened 2019-11-15
Rename FitResult properties

### LINKED ISSUE #266: HoloPy paper uses FitResult.best_fit() as a method
Right now it's a property in the 3.3 branch so the code examples in the HoloPy paper would fail.

I think the best solution is to rename the property and have the method point there.

This also solves the issue @briandleahy has raised that the name best_fit is ambiguous since it could refer to the best fit parameters or the forward model evaluated with best fit parameters.

barkls:
Fixes #266 
A couple notes:

I really struggled with what to call the result of a forward calculation using MAP parameter values from a fitting/MCMC calculation. Previously this was called `best_fit`; the best I could come up with (after brainstorming with @carolinesmartin) now is `inferred_hologram`. This name isn't great because HoloPy supports doing inference with arbitrary forward functions, so then the result of your inference on another scattering calculation would be called a "hologram" which is confusing/wrong. If anyone has a suggestion for a better name, please let me know!

This PR is against the branch in #305 since it builds on those changes. It can be merged before or after that PR, depending on how much revision they each require.


holopy/inference/result.py
```diff
@@ -173,22 +184,35 @@ def _save(self, filename, **kwargs):
...
+        # this method is published in the HoloPy paper
+        from holopy.fitting import fit_warning
+        fit_warning('FitResult.inferred_hologram', 'SamplingResult.best_fit()')
+        return self.inferred_hologram
+
+    def output_scatterer(self):
+        from holopy.fitting import fit_warning
+        fit_warning('FitResult.inferred_scatterer',
+                    'SamplingResult.output_scatterer()')
+        return self.inferred_hologram
```
briandleahy: Shouldn't this return ```SamplingResult.output_scatterer``` rather than ```self.inferred_hologram```?

If so, can we capture this with a test?

holopy/inference/tests/test_result.py
```diff
@@ -71,36 +73,77 @@ def test_properties(self):
...
+    def test_inferred_hologram(self):
+        result = generate_fit_result()
+        self.assertAlmostEqual(
+            result.inferred_hologram.mean().item(), 1.005387, places=6)
+        self.assertTrue(hasattr(result, '_inferred_hologram'))
+
+    @attr("medium")
+    def test_best_fit_returns_inferred_hologram(self):
+        result = generate_fit_result()
+        with warnings.catch_warnings():
```
briandleahy: This test is failing on Travis. I think you need to do:
```python
with warnings.catch_warnings():
    warnings.simplefilter('ignore')
    best_fit = result.best_fit()
```

holopy/inference/result.py
barkls: Good catch!

holopy/inference/tests/test_result.py
barkls: Should be fixed if Travis passes.

barkls:
Thanks for the review! I fixed both issues. I think it's safe to merge this and #305 in either order, so I defer to you on that.

vnmanoharan:
Minor point: do we really need the `inferred_` prefixes?  I understand that you want to distinguish from the `guess_` values but since the object is called `FitResults`, it would probably be clear that, for example, `FitResults.parameters` refers to the results and not the guess.

barkls:
I agree with @vnmanoharan. What do you think @briandleahy?

briandleahy:
> I agree with @vnmanoharan. What do you think @briandleahy?

I am fine with the current version since right now `FitResults` also knows about the initial parameters, but I defer to you two. I think clearer is always better, but clearer is in the eye of the beholder and redundant and clear are not the same thing. So I defer to you two.

vnmanoharan:
In that case, let's remove the `inferred_` prefixes, since they are redundant.

barkls:
Not sure why travis hasn't been triggered. Closing and reopening to see if that helps.

barkls:
Travis still isn't showing the completed build. However, you can [see that it passes](https://travis-ci.org/manoharan-lab/holopy/builds/616928228) by clicking on the "Details" link above.

Ready to merge @briandleahy?

briandleahy:
Yeah I don't know what's going on. I'm going to merge it; we can always un-do / fix it if there is a problem.

PR #311, opened 2019-12-12
Fixscipyleastsq

### LINKED ISSUE #251: scipy optimize strategy returns incorrect uncertainties
I see that scipy.optimize is now used in the 3.3 version. I tested it and the uncertainties given are a bit strange (changing units doesn't change the values). I was wondering if using lmfit instead would be better, it extends many optimization options of scipy.optimize and also add directly a good uncertainties computation.

briandleahy:
Fixes #251 .

@barkls I have 1 question for you on some code in `scipyfit` and `nmpfit`; I've tagged you in a comment below. 

Other than that I'd love to hear review comments.


holopy/inference/scipyfit.py
```diff
@@ -72,6 +72,7 @@ def residual(rescaled_values):
                 parameters, rescaled_values)
             noise = model._find_noise(unscaled_values, data)
             residuals = model._residuals(unscaled_values, data, noise)
+            # FIXME why isn't this just `sqrt(model.lnprior)`?
             prior = np.sqrt(guess_prior - model.lnprior(unscaled_values))
```
briandleahy: Hey @barkls -- do you know why this line is
```python
prior = np.sqrt(guess_prior - model.lnprior(unscaled_values))
```
rather than just
```python
prior = np.sqrt(model.lnprior(unscaled_values)
```
Unless I'm missing something, the current version will (i) push the prior to the guessed values, not to the prior center, and (ii) if `model.lnprior > guess_prior` then there will be `nan`s here. 

The same line is in nmpfit as well.

barkls:
Four points that all apply to both scipyfit and nmpfit:

1) I think this is there to avoid adding really large values that drown out
the rest of the residuals, specifically in the case of a Uniform prior,
where this just removes a constant from all values.

2) The guess value is the prior center in all implemented cases (Gaussian,
Uniform, BoundedGaussian). The calculated lnprior will always be lower than
or equal to the guess prior.

3) I think there's a problem with how the handling of priors are
implemented. I've been experimenting with some changes that I'm not ready
to push, but I believe the sqrt should act on the lnprob for each parameter
and not on the total lnprior. I'm happy to remove this functionality (error
if not all Uniform priors) or release it as is. In either case we'd plan to
have a more rigourously tested implementation for the next release.

4) You've made adjustments to model._residuals that should maybe be
reflected here as well?

briandleahy:
Hey @barkls or @ralex0  -- do either of you know what is up with these failing tests? It looks like its a yaml problem. pyyaml 5.2 was released on Dec 2; I'm not sure if there are bugs with pyyaml 5.2 or if there is holopy code we need to update.

briandleahy:
Thanks for clarifying that. The only thing I am worried about is 
> The guess value is the prior center in all implemented cases (Gaussian,
Uniform, BoundedGaussian). 
If someone decides later to decouple the guess value from the maximum a priori value (e.g. guessing the mean of the prior for a skewed distribution), then `guess_prior - model.lnprior(unscaled_values)` could be negative, and `sqrt(guess_prior - model.lnprior(unscaled_values))` can give `nan`, which would cause all sorts of problems. 

But given your point (3), perhaps I should just leave this alone for now and aim for better not perfect with this PR. I'm happy to either fix it or raise an issue w/ a 3.4 milestone tag if you'd like. 

Also, you state:
> 4) You've made adjustments to model._residuals that should maybe be
reflected here as well?

Good catch! Should be fixed by 7f02687 . I feel we should try to capture this in a unit test, but this is part of the inference api needing better testing and not for 3.3 at this point imo.

barkls:
> But given your point (3), perhaps I should just leave this alone for now and aim for better not perfect with this PR. I'm happy to either fix it or raise an issue w/ a 3.4 milestone tag if you'd like.

I think it's fine for now, since the potentially problematic use-case doesn't exist yet, and we anticipate redoing the current implementation anyway. We should have an issue though.

barkls:
close and reopen to trigger travis

PR #312, opened 2019-12-13
speed up tests

barkls:
Mostly just the super slow one, but I also tinkered with a couple others.

### DIFF
--- a/holopy/inference/tests/test_nmpfit.py
+++ b/holopy/inference/tests/test_nmpfit.py
@@ -213,7 +213,7 @@ def test_serialization():
 
 def test_integer_correctness():
     # we keep having bugs where the fitter doesn't
-    schema = detector_grid(shape = 10, spacing = 1)
+    schema = detector_grid(shape=10, spacing=1)
     s = Sphere(center = (10.2, 9.8, 10.3), r = .5, n = 1.58)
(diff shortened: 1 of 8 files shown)


holopy/inference/tests/test_nmpfit.py
```diff
@@ -213,7 +213,7 @@ def test_serialization():
 
 def test_integer_correctness():
     # we keep having bugs where the fitter doesn't
-    schema = detector_grid(shape = 100, spacing = .1)
+    schema = detector_grid(shape = 10, spacing = 1)
```
briandleahy: Since we're changing this line, can we pep8 it?

barkls:
Test is trivially fast now since it doesn't actually do any sampling. Here we just check that the `sample` function calls `model.sample` with defaults. Actual behaviour of `model.sample` is tested elsewhere.

barkls:
Thanks for fixing that. I'm surprised you can push to my personal branch...

briandleahy:
> Thanks for fixing that. I'm surprised you can push to my personal branch...

BWAHAHAHA :smiling_imp: . 

That's what the "allow edits from maintainers" checkbox. does.

barkls:
did not mean to close...

PR #320, opened 2020-02-18
3.3.2

barkls:
A couple minor patches:

- better handling of missing matplotlib, prompted by conda-forge/holopy-feedstock#27
- fix a typo in docs to help diagnose display issues
- suppress DDA output when running tests

### DIFF
--- a/docs/source/users/install.rst
+++ b/docs/source/users/install.rst
@@ -43,8 +43,8 @@ You will probably be most comfortable using HoloPy in Jupyter (resembles
 Mathematica) or Spyder (resembles Matlab) interfaces. HoloPy is designed to be used with an
 interactive backend. In the console, try running::
 
-    from holopy import test_display
-    test_display()
+    from holopy import check_display
+    check_display()
 
 You should see an image, and you should be able to change
 the square to a circle or diamond by using the left/right arrow keys. If you
--- a/docs/source/users/install.rst
+++ b/docs/source/users/install.rst
@@ -43,8 +43,8 @@ You will probably be most comfortable using HoloPy in Jupyter (resembles
 Mathematica) or Spyder (resembles Matlab) interfaces. HoloPy is designed to be used with an
 interactive backend. In the console, try running::
 
-    from holopy import test_disp
-    test_disp()
+    from holopy import test_display
+    test_display()
 
 You should see an image, and you should be able to change
 the square to a circle or diamond by using the left/right arrow keys. If you
--- a/holopy/core/io/__init__.py
+++ b/holopy/core/io/__init__.py
@@ -18,4 +18,4 @@
 from holopy.core.io.io import (load, load_image, save, save_image,
     get_example_data, get_example_data_path, load_average)
 from holopy.core.io.vis import (show, display_image, show_scatterer_slices,
-                                test_display)
+                                check_display)
--- a/holopy/core/io/__init__.py
+++ b/holopy/core/io/__init__.py
@@ -17,5 +17,5 @@
 # along with HoloPy.  If not, see <http://www.gnu.org/licenses/>.
 from holopy.core.io.io import (load, load_image, save, save_image,
     get_example_data, get_example_data_path, load_average)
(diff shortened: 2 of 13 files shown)


docs/source/users/install.rst
```diff
@@ -43,8 +43,8 @@ You will probably be most comfortable using HoloPy in Jupyter (resembles
 Mathematica) or Spyder (resembles Matlab) interfaces. HoloPy is designed to be used with an
 interactive backend. In the console, try running::
 
-    from holopy import test_disp
-    test_disp()
+    from holopy import test_display
+    test_display()
```
briandleahy: :+1:

holopy/core/tests/test_vis.py
```diff
@@ -244,6 +239,13 @@ def test_maintains_metadata(self):
 class ShowTest(unittest.TestCase):
     @attr("medium")
     def test_show(self):
+        try:
```
briandleahy: My 2 cents:

I'd rather see the import at the top of the file, w/ a `unittest.skipIf(NO_MATPLOTLIB )` or something here. You can even do `from holopy.core.io.vis import _NO_MATPLOTLIB` I believe.

holopy/scattering/theory/dda.py
```diff
@@ -73,11 +73,13 @@ class DDA(ScatteringTheory):
...
     """
     def __init__(self, n_cpu = 1, max_dpl_size=None, use_indicators=True,
-                 keep_raw_calculations=False, addacmd=[]):
+                 keep_raw_calculations=False, addacmd=[],
+                 suppress_C_output=True):
 
         # Check that adda is present and able to run
         try:
-            subprocess.check_call(['adda', '-V'])
+            with SuppressOutput():
```
briandleahy: Thanks for suppressing output on the DDA tests! I meant to do this when I suppressed the output on the tmatrix tests but I forgot to do so.

holopy/core/io/__init__.py
```diff
@@ -17,5 +17,5 @@
 # along with HoloPy.  If not, see <http://www.gnu.org/licenses/>.
 from holopy.core.io.io import (load, load_image, save, save_image,
     get_example_data, get_example_data_path, load_average)
-from holopy.core.io.vis import (show, test_display, display_image,
-                                show_scatterer_slices)
+from holopy.core.io.vis import show, display_image, show_scatterer_slices
+from holopy.core.io.vis import _test_display as test_display
```
briandleahy: Can't we just rename the function in `holopy.core.io.vis` to be `test_display`, instead of calling it `_test_display` and then changing its name on the import here?

holopy/core/io/vis.py
```diff
@@ -268,7 +276,7 @@ def show_scatterer_slices(scatterer, spacing):
     show2d(vol)
 
 
-def test_display():
+def _test_display():
```
briandleahy: (See my comment on `holopy.core.io.__init__`)

holopy/core/io/__init__.py
barkls: it was named test_display before this PR, but nose was picking it up and complaining if there was no matplotlib. This PR originally gave it an underscore, but only in the hp.core.io.vis namespace. Docs point to it in the highest-level holopy namespace (under an incorrect name before this PR). 

I also considered having nose skip or disable it, but it's not really a test so I didn't think that was the right solution.

Now I've renamed it back to its original but have it run a separate matplotlib check that warns instead of raising an exception, so nose will run the code but not complain.

holopy/core/tests/test_vis.py
barkls: agreed this is better. Had to check again though because it needs matplotlib.

holopy/core/io/__init__.py
briandleahy: Ahhh... I see. Can we call it `check_display` then, to avoid nose sniffing it out? Or is it bad form to change this public-facing function on a minor release?

This is a minor issue and an improvement over the old code, so I'm fine with whatever you decide.

Issue #323, opened 2020-03-18
Saving of reconstruction volume plots


MatthiasGi:
To save a hologram plotted by the library one can simply write:
```python
hp.show(holo)
plt.savefig("holo.png")
```

For reconstruction volume this isn't as easy because the images are "hidden" behind keystrokes. I therefore extended `holopy.core.io.vis.py` to offer methods for this use case. A similar thing could be done to `holopy.core.io.io.py` for the `save_image`-method.

Is there any interest in my changes? I'm happy to make a pull request for them. Or are there already good methods to achieve what I want (that simply aren't documented).

barkls:
It's true that there isn't a good way in HoloPy to save a reconstruction stack, and a pull request adding that functionality would be appreciated. Unfortunately the code in `holopy.core.io.vis.py` and `holopy.core.io.io.py` is some of the messiest in HoloPy and should be completely refactored at some point. However, There's no timeline for this, and anything you add would be incorporated into    any rewriting of these files.

briandleahy:
Fixed by #324 and #325

PR #324, opened 2020-03-19
Updatet vis.py to also allow to save generated plots

MatthiasGi:
In reference to #323 I updated the `vis.py` to be able to save generated Plots even for reconstruction volumes.

### DIFF
--- a/holopy/core/io/vis.py
+++ b/holopy/core/io/vis.py
@@ -72,6 +72,48 @@ def show(o, scaling='auto', vert_axis='x', horiz_axis='y',
     else:
         raise VisualizationNotImplemented(o)
 
+def save(o, names, scaling='auto', vert_axis='x', horiz_axis='y',
+                    depth_axis='z', colour_axis='illumination'):
+    """
+    Saves a hologram or reconstruction to (a) file(s).
+
+    Parameters
+    ----------
+    o : xarray.DataArray or ndarray
+        Object to visualize.
+    names : str
+        Name(s) of the file(s). If there is only one image contained (e.g.
+        hologram), the name will be used as a file name. If o contains more
+        plottable images (e.g. reconstruction volume), this string will be
+        formatted to get a counting index inserted. Example:
+        "reconstruction-%i.png"
+    scaling : (float, float), optional
+        (min, max) value to display in image, default is full range of o.
+    vert_axis : str, optional
+        axis to display vertically, default x.
+    horiz_axis : str, optional
+        axis to display horizontally, default y.
+    depth_axis : str, optional
+        axis to scroll with arrow keys, default 'z'.
+    colour_axis : str, optional
+        axis to display as RGB channels in colour image, default illumination.
+
+    Notes
+    -----
+    Loads plotting library the first time it is required (so that we don't have
+    to import all of matplotlib or mayavi just to load holopy)
+    """
+    if isinstance(o, (xr.DataArray, np.ndarray, list, tuple)):
+        im = display_image(o, scaling, vert_axis, horiz_axis, depth_axis,
+                           colour_axis)
+        s = Show2D(im)
+        if len (im) > 1:
+            s.saveAll(names)
+        else:
+            s.save(names)
+    else:
+        raise VisualizationNotImplemented(o)
+
 
 class Show2D(object):
     def __init__(self, im):
@@ -143,6 +185,32 @@ class Show2D(object):
             dimname = self.im.dims[0]
             titlestring = '{} = {}'.format(dimname, self.im[dimname].values[self.i])
             self.ax.set_title(titlestring)
+
+    def save(self, name):
+        """
+        Saves the currently displayed Plot into a file.
+
+        Parameters
+        ----------
+        name : str
+            Name for the file to save to.
+        """
+        self.draw()
+        self.fig.savefig(name)
+
+    def saveAll(self, names):
+        """
+        Saves the complete stack of images into separate files.
+
+        Parameters
+        ----------
+        names : str
+            Name of the files, will be formatted to get a counting index
+            inserted. Example: "reconstruction-%i.png"
+        """
+        for i in range(len(self.im)):
+            self.i = i
+            self.save(names % i)
 
 
 def display_image(im, scaling='auto', vert_axis='x', horiz_axis='y',
@@ -293,4 +361,3 @@ def test_display():
                 a[50-i, 50-j, 1] = 1
                 a[50+i, 50-j, 1] = 1
     show(a)
-
--- a/holopy/core/io/vis.py
+++ b/holopy/core/io/vis.py
@@ -72,6 +72,7 @@ def show(o, scaling='auto', vert_axis='x', horiz_axis='y',
     else:
         raise VisualizationNotImplemented(o)
 
+
 def save_plot(filenames, data, scaling='auto', vert_axis='x', horiz_axis='y',
               depth_axis='z', colour_axis='illumination'):
     """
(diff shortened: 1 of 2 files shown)


holopy/core/io/vis.py
```diff
@@ -72,6 +72,48 @@ def show(o, scaling='auto', vert_axis='x', horiz_axis='y',
     else:
         raise VisualizationNotImplemented(o)
 
+def save(o, names, scaling='auto', vert_axis='x', horiz_axis='y',
```
briandleahy: Can we change the function name to a more descriptive name? Perhaps `save_reconstruction` or `save_plotted_hologram`?

holopy/core/io/vis.py
```diff
@@ -72,6 +72,48 @@ def show(o, scaling='auto', vert_axis='x', horiz_axis='y',
...
+                    depth_axis='z', colour_axis='illumination'):
+    """
+    Saves a hologram or reconstruction to (a) file(s).
+
+    Parameters
+    ----------
+    o : xarray.DataArray or ndarray
```
briandleahy: I realize you are copying the syntax from `holopy.core.io.vis.show`, but can we change the variable name from `o` to something more descriptive? Perhaps `holograms` or `data`?

holopy/core/io/vis.py
```diff
@@ -72,6 +72,48 @@ def show(o, scaling='auto', vert_axis='x', horiz_axis='y',
...
+        Object to visualize.
+    names : str
+        Name(s) of the file(s). If there is only one image contained (e.g.
+        hologram), the name will be used as a file name. If o contains more
+        plottable images (e.g. reconstruction volume), this string will be
+        formatted to get a counting index inserted. Example:
+        "reconstruction-%i.png"
```
briandleahy: As I said regarding #323, I personally think a list-like of filenames is better than a format string.

holopy/core/io/vis.py
```diff
@@ -144,6 +186,32 @@ def _title(self):
...
+        """
+        Saves the currently displayed Plot into a file.
+
+        Parameters
+        ----------
+        name : str
+            Name for the file to save to.
+        """
+        self.draw()
+        self.fig.savefig(name)
```
briandleahy: :+1:

holopy/core/io/vis.py
```diff
@@ -144,6 +186,32 @@ def _title(self):
...
+
+        Parameters
+        ----------
+        names : str
+            Name of the files, will be formatted to get a counting index
+            inserted. Example: "reconstruction-%i.png"
+        """
+        for i in range(len(self.im)):
+            self.i = i
+            self.save(names % i)
```
briandleahy: :+1:

holopy/core/io/vis.py
```diff
@@ -144,6 +186,32 @@ def _title(self):
...
+
+        Parameters
+        ----------
+        name : str
+            Name for the file to save to.
+        """
+        self.draw()
+        self.fig.savefig(name)
+
+    def saveAll(self, names):
```
briandleahy: per pep8, can we call this method `save_all` instead of `saveAll`?

holopy/core/io/vis.py
```diff
@@ -72,6 +72,48 @@ def show(o, scaling='auto', vert_axis='x', horiz_axis='y',
...
+    scaling : (float, float), optional
+        (min, max) value to display in image, default is full range of o.
+    vert_axis : str, optional
+        axis to display vertically, default x.
+    horiz_axis : str, optional
+        axis to display horizontally, default y.
+    depth_axis : str, optional
+        axis to scroll with arrow keys, default 'z'.
+    colour_axis : str, optional
+        axis to display as RGB channels in colour image, default illumination.
```
briandleahy: If you can think of a pithy example, that would be great to add here!

holopy/core/io/vis.py
MatthiasGi: As the method is able to save holograms and reconstruction, I relabeled it to `save_plot`. Is this better?

holopy/core/io/vis.py
MatthiasGi: Done!

holopy/core/io/vis.py
MatthiasGi: Made it!

holopy/core/io/vis.py
MatthiasGi: Thanks! 😄

holopy/core/io/vis.py
MatthiasGi: Are you thinking of an example for the parameters `depth_axis` and `colour_axis` or the method itself?

holopy/core/io/vis.py
MatthiasGi: Additional comment: Shouldn't `show` then be renamed, too? As they process the same type of data?

MatthiasGi:
I implemented a starting point for a test. What's missing: Comparing the produced image with what's expected. I created the method `_compare_twp_images_by_fname` for that. Maybe someone has the right code snippet at hand?

MatthiasGi:
**Additional comment:** Unfortunately I wasn't able to get this not compiled version of HoloPy running. Thus I'm not able to really test what I'm writing. I'm developing code snippets outside of the library and then trying to insert them. This is the reason for typing errors. Please excuse them and be extra careful!

holopy/core/tests/test_vis.py
```diff
@@ -64,6 +65,37 @@ def convert_ndarray_to_xarray(array, extra_dims=None):
...
+        save_plot(self.filenames[0], self.holograms[0])
+        show(self.holograms[0])
+        fname = os.path.join(self.tempdir, 'test.png')
+        plt.savefig(fname)
+        plt.clf()
+        self._compare_two_images_by_fname(self.filenames[0], fname)
+
+    @attr("fast")
+    def test_save_multiple_images(self):
+        save_plot(self.filenames, self.holograms)
```
briandleahy: Hey @MatthiasGi , 

When I clone this branch and run this code on my machine, the ```save_plot``` line crashes with an error:
```python
BadImage: Too many dims on ndarray to output properly.
```
The reason is that the loaded images end up as shape [2, 1, N, N] rather than as 3D images. So I'm a little confused as to exactly what you are trying to do here. Does this code run on your machine? If you can help me out a bit understanding what you are trying to do, I can probably help you fix this and merge it in.

barkls: I believe the problem is that `self.holograms` is a list of xr.DataArray objects. The `hp.core.io.vis.display_image` function called in `save_plot` is designed to take holograms in a variety of input formats and output a standardized xr.DataArray with known dimension order for saving or displaying. A list of xr.DataArrays is not something it knows how to handle and I think it's getting cast to numpy ndarray.

I think the best solution might be to use `xr.concat` or `hp.core.metadata.clean_concat` to construct the test stack instead of packing as a list.

briandleahy: Awesome, thanks @barkls. I fixed this in the next 3 commits. If you could quickly check that I am clean_concat'ing the right way, that would be *awesome*.

briandleahy:
Hey @MatthiasGi , I made some other minor changes here to fix the failing tests. Can you take a look at the changes and make sure that the still capture the spirit of your PR? If so, then once the tests pass on Travis we'll merge this in. 

Thanks!

MatthiasGi:
Hi @briandleahy! Thanks for your update – somehow I didn't receive a notification and didn't notice your reply. I'm not seeing anything that changed my intent in creating the PR. Thanks a lot for your assistance!

PR #325, opened 2020-03-20
Quick hack for saving reconstruction volumes

MatthiasGi:
As referenced in #323 I wrote a quick hack to be able to save images from a reconstruction volume. There are two things to consider:

1. Maybe the created `_save_image`-method can be used for the existing `save_image`. The only difference is the `im = im.values` (note the missing index of 0) in the new method. Maybe there is some clever if-construct that is able to handle this automatically? I'm unfortunately not familiar with `xarray` at all. Maybe someone can refactor the code?

2. `save_images` has no check if the given data has the correct format. I think this is acceptable for a quick hack but not for a robust library. Maybe someone can look into this – I'm still not familiar with `xarray`.

### DIFF
--- a/holopy/core/io/io.py
+++ b/holopy/core/io/io.py
@@ -323,12 +323,12 @@ def save_image(filename, im, scaling='auto', depth=8):
 
     Parameters
     ----------
-    im : ndarray or :class:`holopy.image.Image`
-        image to save.
     filename : basestring
         filename in which to save image. If im is an image the
         function should default to the image's name field if no
         filename is specified
+    im : ndarray or :class:`holopy.image.Image`
+        image to save.
     scaling : 'auto', None, or (Int, Int)
         How the image should be scaled for saving. Ignored for float
         output. It defaults to auto, use the full range of the output
@@ -380,19 +380,19 @@ def save_image(filename, im, scaling='auto', depth=8):
         pilimage.fromarray(im).save(filename)
 
 
-def save_images(ims, filenames, scaling='auto', depth=8):
+def save_images(filenames, ims, scaling='auto', depth=8):
     """
     Saves a volume as separate images (think of reconstruction volumes).
 
     Parameters
     ----------
-    ims : ndarray or :class:`holopy.image.Image`
-        Images to save, with separate z-coordinates from which they each will
-        be selected.
     filenames : str
         List of filenames. There have to be the same number of filenames as of
         images to save. Each image will be saved in the corresponding file with
         the same index.
+    ims : ndarray or :class:`holopy.image.Image`
+        Images to save, with separate z-coordinates from which they each will
+        be selected.
     scaling : 'auto', None, or (Int, Int)
         How the images should be scaled for saving. Ignored for float output.
         It defaults to auto, use the full range of the output format. Other
@@ -412,18 +412,18 @@ def save_images(ims, filenames, scaling='auto', depth=8):
     for i, im in enumerate(ims): _save_im(im, filenames[i], depth)
 
 
-def _save_im(im, filename, depth=8):
+def _save_im(filename, im, depth=8):
     """
     Internal single-image-save-method to be used in save_images. Maybe it can
     be merged with save_image.
 
     Parameters
     ----------
-    im : ndarray or :class:`holopy.image.Image`
-        Image to save.
     filename : basestring
         Filename in which to save image. If im is an image the function should
         default to the image's name field if no filename is specified
+    im : ndarray or :class:`holopy.image.Image`
+        Image to save.
     depth : 8, 16 or 'float'
         What type of image to save. Options other than 8bit may not be
         supported for many image types. You probably don't want to save 8bit
--- a/holopy/core/io/io.py
+++ b/holopy/core/io/io.py
@@ -379,6 +379,94 @@ def save_image(filename, im, scaling='auto', depth=8):
     else:
         pilimage.fromarray(im).save(filename)
 
+def save_images(filenames, ims, scaling='auto', depth = 8):
+    """
+    Saves a volume as separate images (think of reconstruction volumes).
+
+    Parameters
+    ----------
+    ims : ndarray or :class:`holopy.image.Image`
+        Images to save, with separate z-coordinates from which they each will be
+        selected.
+    filename : str
+        Filename in which to save image. Has to contain a format placeholder for
+        the z-values. E.g. "testfile%f.png"
+    scaling : 'auto', None, or (Int, Int)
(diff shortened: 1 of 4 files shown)


briandleahy:
Hey @MatthiasGi , thanks for doing this! 

Maybe I'm just dense, but can you explain the difference between #324 and #325? It seems that both PRs provide a way to save a stack of holograms to a file, but #325 provides more flexibility in dealing with different bit depths and more power in packing metadata into the tiff exif. Do you want to include both of them in holopy? Or were you showing two different ways to achieve the same end?

After that I'll do a quick code review then we'll merge this (or these) in!

Thanks!
@briandleahy

MatthiasGi:
Hi @briandleahy! The difference is in what files are generated. #325 generates only the pure images (as `save_image` does for one single image of a hologram). #324 instead saves whole plots with axis and so on as they are displayed by `hp.show(rec_vol)`. If the difference isn't still clear (maybe I'm articulating myself in a weird way), I can add images to make it more clear.

holopy/core/io/io.py
```diff
@@ -380,6 +380,96 @@ def save_image(filename, im, scaling='auto', depth=8):
...
+    ims = display_image(ims, scaling)
+    for im in ims:
+        z = im.z.values.item(0)  # TODO: There must be a nicer way…
+        _save_im(filenames % z, im, scaling, depth)
+
+
+def _save_im(filename, im, scaling='auto', depth=8):
+    """
+    Internal single-image-save-method to be used in save_images. Maybe it can
+    be merged with save_image.
```
briandleahy: it looks like the original save_image shares a lot of functionality with this, so I think you're right that it can be merged with save image. 

If you are up to the challenge, go ahead! If not, don't worry -- I'll make a separate issue and one of me, @barkls , or @ralex0 will take care of it quickly.

holopy/core/io/io.py
```diff
@@ -380,6 +380,96 @@ def save_image(filename, im, scaling='auto', depth=8):
         pilimage.fromarray(im).save(filename)
 
 
+def save_images(filenames, ims, scaling='auto', depth=8):
+    """
+    Saves a volume as separate images (think of reconstruction volumes).
+
+    Parameters
+    ----------
+    ims : ndarray or :class:`holopy.image.Image`
```
briandleahy: The function argument takes parameters in the order (filenames, ims), but the docstring provides them in order (ims, filenames). Can you switch the order to be consistent with the function argument order? I see that you just copied the docstring for save_image, which has this order. So if you think this is out of the scope of this PR, let me know and I'll add it to a TODO list for later.

holopy/core/io/io.py
briandleahy: it looks like the original save_image shares a lot of functionality with this, so I think you're right that it can be merged with save image.

If you are up to the challenge, go ahead! If not, don't worry -- I'll make a separate issue and one of me, @barkls , or @ralex0 will take care of it quickly.

holopy/core/io/io.py
```diff
@@ -380,6 +380,96 @@ def save_image(filename, im, scaling='auto', depth=8):
...
+        Images to save, with separate z-coordinates from which they each will
+        be selected.
+    filename : str
```
briandleahy: In my personal opinion, I think a better practice would be to pass in an iterable of filenames, rather than a format string. I think the format string is a little less intuitive and creates more potential for problems.

holopy/core/io/io.py
```diff
@@ -380,6 +380,96 @@ def save_image(filename, im, scaling='auto', depth=8):
...
+    scaling : 'auto', None, or (Int, Int)
+        How the images should be scaled for saving. Ignored for float output.
+        It defaults to auto, use the full range of the output format. Other
+        options are None, meaning no scaling, or a pair of integers specifying
+        the values which should be set to the maximum and minimum values of the
+        image format.
+    depth : 8, 16 or 'float'
+        What type of image to save. Options other than 8bit may not be
+        supported for many image types. You probably don't want to save 8bit
+        images without some kind of scaling.
```
briandleahy: If you can think of a pithy minimal-working example here for the docs, that would be awesome. If not, no worries.

holopy/core/io/io.py
MatthiasGi: Done!

holopy/core/io/io.py
MatthiasGi: `save_image` could be rewritten to:
```python
def save_image(filename, im, scaling='auto', depth=8):
    im = display_image(im, scaling)
    _save_im(filename, im, depth)
```

**But**: There is the line `im = im.values` in `_save_im` which in `save_image` is `im = im.values[0]`. As I'm not familiar with the structure of the image-object, someone would need to abstract this line. I could only introduce a boolean parameter flag, split the method up or pass in the values as additional parameter. Each method doesn't seem that great to me. Maybe there is a simple way to check, if there is the need to access the item 0?

holopy/core/io/io.py
MatthiasGi: Done!

holopy/core/io/io.py
MatthiasGi: An example for the depth-Parameter? I'm not sure – as I just copied it other. Or of the whole method itself?

holopy/core/io/io.py
briandleahy: Let's leave it for a future refactor of io.py then.

holopy/core/io/io.py
briandleahy: Sorry! I meant an example of the whole method itself.

holopy/core/io/io.py
barkls: The `im.values[0]` call is specifically to return only the first slice of a volume reconstruction. This is exactly the problem you are trying to fix with this PR!

holopy/core/io/io.py
MatthiasGi: Okay, no problem. In the actual docs or inside of this comment?

holopy/core/io/io.py
MatthiasGi: Thanks for the clarification @barkls! I tried to make a nice adjustment of the method. Please check it out!

briandleahy:
Hi @MatthiasGi 

These commits should fix the failing tests in #325. I took the liberty of pushing to your branch since allowed commits from maintainers; I hope you don't mind. 

Once the checks pass on Travis, I'll merge it in, and then I'll address #324

Thanks!

MatthiasGi:
Of course I don’t mind! Thanks for cleaning up where I left a mess!

briandleahy:
Closing and re-opening to trigger Travis

PR #328, opened 2020-04-03
Fixfft

briandleahy:
This fixes a bug that appeared when scipy upgraded to 1.4.0. The bug arose because holopy did fft's on xarrays, which do not share all the attributes of numpy ndarrays (despite subclassing from them...). I fixed this by always doing the fft on the numpy array, then packing the result back into the xarray (this functionality was already present). 

Since I was touching the fourier.py file, I went ahead and brought us into the 2010s and used the  numpy.fft module instead of scipy.fftpack, which has been legacy for a while and has been responsible for some trouble in holopy. As a result, I deprecated the `overwrite` kwarg for fft's, which was performance based but not actually used for performance in the code (it was called, but the in-place modification did not matter due to the nature of the call.) Calling `hp.core.proces.fourier.fft` or `hp.core.process.fourier.ifft` with the `overwrite=True` now warns the user. If you want, I can raise an issue that this get fixed for the next minor release. Or, if this is deemed internal, I can just remove the overwrite kwarg.

This is based off of #327 , so that should be merged in first.


barkls:
Thanks for doing this, and for adding  some tests!

I think it's fine to remove the "overwrite" keyword. It looks like there's merge conflict in .travis.yml since #327 already pinned xarray.

briandleahy:
Thanks @barkls . I completely removed the overwrite kwarg in 210a006. I also did a quick grep and removed fftpack from throughout the rest of holopy (it only occurred in a few places).

Let me know if there is something else I can improve.

briandleahy:
Oh! and I fixed the .travis.yml merge conflict :smile:

barkls:
Looks good to me!

PR #329, opened 2020-04-06
Branch 3.3.2

barkls:
Let's do a mini release so that we can get dependencies (scipy, xarray, etc.) cleaned up on conda-forge.


briandleahy:
Sounds good to me. I think we should leave the pep8speaks comments alone for internal merge PRs as well. 

There are some easy docs issues open, but they don't need to be included in this mini-release. So I'll approve and merge.

PR #330, opened 2020-04-06
bump develop version number

barkls:
Branch protections mean I can't push directly...we might want to adjust that if it becomes tedious to open many trivial PRs.


briandleahy:
The administrator should be able to still merge directly. 

Regardless, I think it's OK for now, since I don't think we'll have that many trivial PRs.

PR #332, opened 2020-04-13
Fixxarraytest

briandleahy:
Hey @barkls , I want some advice from you.

I took a look into the test that fails when xarray.__version__ is 0.15.1 instead of 0.15.0. It turns out that the test fails for a more-or-less silly reason. There is a seemingly unnecessary function ```fix_flat``` defined and only used in `holopy.inference.tests.test_nmpfit`. The function `fix_flat` tries to assign values to an xarray attribute, which is no longer allowed, and which is why the test failed. 

However, for most of what was actually tested (whether fitting with a random subset produces the correct answer), the `fix_flat` call was unnecessary. I deleted the `fix_flat` function and split the test up into several tests; the tests testing whether random subset fitting returns a correct fit pass. 

But, there is another wrinkle -- the original test also tested whether saving the object and loading it again returned an equal object (through `assert_read_matches_write`). It turns out that the random subset data does not save correctly (its attributes changes on being saved & reloaded). The original test (accidentally?) got around this by calling fix_flat, which allowed the result to be saved correctly. However, `fix_flat` is not part of the holopy API, so this test should fail, because right now a user saving a random subset fit will get weird attributes on loading it back in. 

In other it seems that the problem has nothing to do with xarray changes and is actually that the random subset result is not writable natively. (This smells of the problems present in issue #230 ). Right now I have this captured as a skipped failing test. I see 2 ways we could proceed:

* We declare that saving random subset fits is not well supported, and that there is no problem. 
* We say that this is a known bug and raise an issue, which hopefully gets touched when someone standardizes the xarray usage in holopy.

What do you think the best course of action is? One of the above, or something totally different?


barkls:
When you save a `FitResult` object, there's a bunch of code in [`_serialization_ds`](https://github.com/manoharan-lab/holopy/blob/d7f6f4295e6249462dd242734f52903df9c6a502/holopy/inference/result.py#L154) that gets called to massage various parts of the FitResult into something that saves nicely. One issue is the stacked multi-index "flat" coordinate, which gets converted into a simpler "point" coordinate (I don't remember why this was necessary - it might not be anymore). HoloPy should  treat "point" and "flat" coordinates as interchangeable. The `fix_flat` function was cheating in a sense by letting the object pass a save-and-reload test by using our knowledge that "point" and "flat" coordinates should be treated as the same thing. This is related to  #230, but I think it falls more into my first big picture 3.4 goal. Unfortunately that's not on github, but it reads:

barkls:
**Big picture**: We should be able to save & reload any possible holopy object without losing information.

HoloPy stores information in two distinct ways - numerical data in the form of xarray DataArrays, and metadata/procedure objects in the form of hierarchical HoloPyObjects. Right now we save DataArrays as .h5 files, but it always feels a bit like forcing a square peg into a round hole since so much massaging is required to pack/unpack holograms, and it also suffers from the fact that objects that seem like they should behave the same way often don't (Brian's point 2). We serialize HoloPyObjects to strings with yaml, which has its own issues since yaml standards seem to shift not infrequently. We have real problems with objects that combine both forms of data storage, notably FitResult objects, which cause frequent bugs with saving/loading and have accumulated a lot of messy code to help, but not fix the problem.  

Steps toward this that are accomplishable with a 3.4 release are:
1) clean up the xarray situation so that all DataArrays will behave the same way.
2) Rewrite our basic serialization procedure to handle DataArrays as well as HoloPyObjects - either one can contain the other hierarchically. FitResult sort of does this already, but it should be generalized so it's less sensitive to edge cases.

briandleahy:
@barkls  see 433d0e1 (failing test) vs 3c03c6a (test should pass) -- it seems that whatever cma is running, it is not terribly reproducible across systems. I don't know that this means we should do anything differently, just want to point it out.

PR #333, opened 2020-04-15
Lens

ralex0:
This pull request is for the new `Lens` object that can add the effect of a perfect lens to any `ScatteringTheory`. I've included some tests. I have tried this with `Mie`, `Multisphere`, `Tmatrix` and `DDA` and all seem to work. Let me know your thoughts so we can incorporate it into 3.4.

I'm awaiting the PEP8 linter's judgement, so I'll fix whatever it wants later.


holopy/scattering/tests/test_lens.py
```diff
@@ -0,0 +1,213 @@
...
+        integral = np.sum(func(pts) * wts)
+        expected_val = np.sin(LENSMIE.lens_angle) # analytic result
+        assert_allclose(integral, expected_val)
+
+    def test_integrate_over_phi_with_quad_points(self):
+        pts = LENSMIE._phi_pts
+        wts = LENSMIE._phi_wts
+        func = lambda x: np.cos(np.pi * x)
+        integral = np.sum(func(pts) * wts)
+        expected_val = np.sin(2 * np.pi ** 2) /  np.pi # analytic result
```
briandleahy, line 53: Wait I'm confused. The integral of cos(pi*x) from 0 to 2pi is sin(2*pi^2) / pi? Shouldn't it be zero?

If it is zero, that may be a sign that the default tolerances in `assert_allclose` are too soft; it might be good to tighten them.

If I'm just confused, ignore me :smile:

holopy/scattering/tests/test_lens.py
```diff
@@ -0,0 +1,213 @@
...
+        medium_wavevec = 2 * np.pi / test_common.wavelen
+        medium_index = test_common.index
+        illum_polarization = test_common.xpolarization
+
+        theory_old = MieLens(lens_angle=LENS_ANGLE)
+        theory_new = LENSMIE
+
+        fields_old = theory_old.calculate_scattered_field(scatterer, detector)
+        fields_new = theory_new.calculate_scattered_field(scatterer, detector)
+
```
briandleahy, line 159: I think the above tests are good, and you should keep them. 

But I think we also need a test that MieLens() and Lens(MIe) are the same at much tighter tolerances, say at 1e-12 levels. This might require increasing the number of quadrature points by a lot for both, so I don't think this is a test that we want to run over a full 64x64 hologram; we should instead test it over a few points (say 5 randomly chosen points at some random z).

The reason that I think this is important is that there are some "small signal" physical effects, which either Lens(Mie) or MieLens could be getting backwards. For instance, there is a cos(2phi) term in the hologram from a sphere, but it is only noticeable at about the 1e-3 level (depending on parameters). So it's possible for one of the two theories to have a sign error and for these tests to still pass.

holopy/scattering/theory/lens.py
```diff
@@ -0,0 +1,192 @@
...
+
+    @classmethod
+    def _integrand_prefactor(cls, sinth, costh, phi, dth, dphi,
+                             krho_p, phi_p, kz_p):
+        if NUMEXPR_INSTALLED:
+            prefactor = ne.evaluate(cls.numexpr_integrand_prefactor1)
+            prefactor *= ne.evaluate(cls.numexpr_integrand_prefactor2)
+            prefactor *= ne.evaluate(cls.numexpr_integrand_prefactor3)
+        else:
+            prefactor = np.exp(1j * krho_p * sinth * np.cos(phi - phi_p))
```
briandleahy, line 126: Is it possible to test that the two pathways produce the same result, with and without numexpr?

briandleahy: alternatively you could use an eval() statement, but I don't really like that option....

holopy/scattering/theory/lens.py
```diff
@@ -0,0 +1,192 @@
...
+        S = np.swapaxes(S, 0, 1)
+        S1 = S[:, :, 1, 1].reshape(self.quad_npts_theta, self.quad_npts_phi, 1)
+        S2 = S[:, :, 0, 0].reshape(self.quad_npts_theta, self.quad_npts_phi, 1)
+        S3 = S[:, :, 0, 1].reshape(self.quad_npts_theta, self.quad_npts_phi, 1)
+        S4 = S[:, :, 1, 0].reshape(self.quad_npts_theta, self.quad_npts_phi, 1)
+        return S1, S2, S3, S4
+
+    @classmethod
+    def _integrand_prll(cls, prefactor, cosphi, sinphi, S1, S2, S3, S4):
+        if NUMEXPR_INSTALLED:
```
briandleahy, line 151: (same comment as above)

holopy/scattering/theory/lens.py
```diff
@@ -0,0 +1,192 @@
...
+            integrand_l = ne.evaluate(cls.numexpr_integrandl)
+        else:
+            integrand_l = prefactor * (cosphi * (cosphi * S2 + sinphi * S3)
+                                     + sinphi * (cosphi * S4 + sinphi * S1))
+        return integrand_l
+
+    @classmethod
+    def _integrand_perp(cls, prefactor, cosphi, sinphi, S1, S2, S3, S4):
+        if NUMEXPR_INSTALLED:
```
briandleahy, line 160: (same comment as above)

holopy/scattering/tests/common.py
```diff
@@ -26,7 +26,7 @@
 pixel_scale = [.1151e-6, .1151e-6]
 index = 1.33
 
-xschema = update_metadata(detector_grid(shape = 128, spacing=pixel_scale), illum_wavelen=wavelen, medium_index=index, illum_polarization=xpolarization)
+xschema = update_metadata(detector_grid(shape=64, spacing=pixel_scale), illum_wavelen=wavelen, medium_index=index, illum_polarization=xpolarization)
```
barkls, line 29: Is this the cause of the failing tests?

holopy/scattering/theory/lens.py
```diff
@@ -0,0 +1,192 @@
+import numpy as np
+
+try:
+    import numexpr as ne
+    NUMEXPR_INSTALLED = True
+except ModuleNotFoundError:
+    NUMEXPR_INSTALLED = False
```
barkls, line 7: It might be nice to get a warning that installing numexpr might speed things up. Right now there's no way to know it's even an option since it silently switches to a fallback.

holopy/scattering/theory/lens.py
```diff
@@ -0,0 +1,192 @@
...
+        xyz = np.zeros([3, prll_component.size], dtype='complex')
+        for i in range(2):
+            xyz[i, :] += prll_component * parallel[i]
+            xyz[i, :] += perp_component * perpendicular[i]
+        return xyz
+
+    def _compute_field_prefactor(self, scatterer, medium_wavevec):
+        return -1. * np.exp(1j * medium_wavevec * scatterer.center[2])
+
+    def _raw_scat_matrs(self, *args, **kwargs):
```
barkls, line 182: I don't like that it passes raw scattering matrices through without modification by the lens. I would prefer raising an error that tells you not to use a lens if you are interested in scattering matrices.

holopy/scattering/theory/lens.py
briandleahy: :+1:

holopy/scattering/theory/lens.py
briandleahy: :+1:

holopy/scattering/tests/common.py
ralex0: Yes, but reverting this change breaks the raw fields tests in test_lens. I'll just make them not use the same detector for now until I sort this out.

holopy/scattering/theory/lens.py
ralex0: Okay I'll add a warning in the next revision.

holopy/scattering/theory/lens.py
ralex0: I'll just remove this and the Theory will automatically inherent this method from parent which the Lens(Theory) was instantiated with.

holopy/scattering/theory/lens.py
ralex0: Yes, I added a few tests that do this. It's a little tricky to do the numpy part without the eval so I had to hard code the numpy expressions into the test. Basically I check that the numexpr string really do what they say the do.

holopy/scattering/tests/test_lens.py
ralex0: I have resolved everything except this comment. It’s not so simple as adjusting test parameters to be 5 pixels and “enough” quadrature points since the numpy guassian quadrature algotithm takes time too.

1000x1000 hologram + 100x100 quadrature takes less way less time than 100x100 hologram + 1000x1000 quadrature.

ralex0:
See my comment in the unresolved thread above

holopy/scattering/tests/test_lens.py
briandleahy: Hmm... how much does the deviation between Lens(Mie) and MieLens() improve if you do (say) 200x200 quadrature?

I think it's important to have some sort of test that at least checks that the deviation decreases between the two equivalent theories as the number of quad pts increases. We don't have to go down to 1e-13 tolerances, but something a little tighter than 5e-3 -- something like 1e-5 or 1e-6 should be fine.

If you want, I can take a look at it. Just let me know; I should have some time tomorrow.

ralex0: Going from 100x100 pixels to 200x200 gives less than x2 of precision but takes more than x2 time. Try it out tomorrow and you can see what I mean.

briandleahy: OK... so I tested whether MieLens() and Lens(Mie()) go to the same values as the number of quadrature points gets large, and the results are weird. 

Basically, both Lens(Mie) and MieLens() converge very rapidly at small distances as the number of quadrature points is increased. However, they do not converge to the same values. When the quadrature points of both are large, the fields differ by ~6e-5 (cf. the new test in e68f0ea ). This is a weird number. 6e-5 is smaller than any physics, but much bigger than numerics should be. And while the difference in the fields is not random, there is no obvious pattern in the differences either (e.g. it's not cos(2phi) symmetry). 

So... I don't know what to do. Right now I'm thinking leave the failing test in there, skipped, merge this in, and raise an issue that there is a subtle problem. My hunch is that the difference is in the computation of the Mie scattering coefficients....

Thoughs @ralex0  and @barkls ?

barkls: I worry about leaving in skipped failing tests because that seems a bit like pushing the problem onto some future HoloPy developer who probably doesn't care too much about that specific issue. I would prefer to have a passing test (with more forgiving tolerance), and raise an issue that it should be looked into. That way there is still a record of the problem and that it needs to be fixed, but we are running the test in the meantime.

barkls:
@briandleahy still has an outstanding changes requested in a review. Dismissing so I can merge, since he confirmed everything looks good in a meeting last week.

briandleahy:
Thanks @barks; I thought I dismissed my outstanding changes earlier.

PR #334, opened 2020-04-17
update to get show_scatterer_slices to work

anna-wang:
show_scatterer_slices wasn't working. update includes:
- changing voxelated scatterer from ndarray to xarray
- removing the rescaling by vmin and vmax for voxelated scatterers. This rescaling causes the visualisation to show incorrectly because the same vmin and vmax are used again in imshow.

### DIFF
--- a/holopy/core/io/vis.py
+++ b/holopy/core/io/vis.py
@@ -129,7 +129,7 @@ class Show2D(object):
 
         self.i = 0
         vmin, vmax = im.attrs['_image_scaling']
-        if im.ndim == 3:
+        if im.ndim == 3 and not hasattr(im,'voxel'):
             #greyscale image
             self.im = im * (vmax-vmin) + vmin
         else:
@@ -341,7 +341,7 @@ def show_scatterer_slices(scatterer, spacing):
         voxel spacing for the visualization
     """
     vol = scatterer.voxelate(spacing, 0)
-    show2d(vol)
+    Show2D(vol)
 
 
 def check_display():
--- a/holopy/core/io/vis.py
+++ b/holopy/core/io/vis.py
@@ -341,7 +341,7 @@ def show_scatterer_slices(scatterer, spacing):
         voxel spacing for the visualization
     """
     vol = scatterer.voxelate(spacing, 0)
-    Show2D(vol)
+    Show2D(display_image(vol))
 
 
 def check_display():
--- a/holopy/scattering/scatterer/scatterer.py
+++ b/holopy/scattering/scatterer/scatterer.py
@@ -99,7 +99,7 @@ class Scatterer(HoloPyObject):
         index = np.ones_like(domains, dtype=dtype) * background
         for i, n in enumerate(ns):
             index[domains==i+1] = n
-        return xr.DataArray(index, attrs = {'_image_scaling':[0,self.n.max()], 'voxel':1})
+        return index
 
     @property
     def guess(self):
--- a/holopy/scattering/scatterer/scatterer.py
+++ b/holopy/scattering/scatterer/scatterer.py
@@ -99,7 +99,7 @@ class Scatterer(HoloPyObject):
         index = np.ones_like(domains, dtype=dtype) * background
         for i, n in enumerate(ns):
             index[domains==i+1] = n
-        return index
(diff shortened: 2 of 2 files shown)


holopy/scattering/scatterer/scatterer.py
```diff
@@ -99,7 +99,7 @@ def index_at(self, points, background=0):
         index = np.ones_like(domains, dtype=dtype) * background
         for i, n in enumerate(ns):
             index[domains==i+1] = n
-        return index
+        return xr.DataArray(index, attrs = {'_image_scaling':[0,self.n.max()], 'voxel':1})
```
barkls, line 102: The `self.n.max()` call is tripping up a few unittests on travis. Specifically, it looks like n is sometimes defined as a float, list, or tuple in tests so the max call is failing. Wrapping `self.n` in `hp.core.metadata.ensure_array` should solve.

barkls, line 102: This is a bit of a missed opportunity to take full advantage of xarray by keeping track of the coordinates of each voxel. It should be pretty straightforward to pass in coords and dims according to the [xarray documentation](http://xarray.pydata.org/en/stable/data-structures.html#creating-a-dataarray). Let me know if you want help implementing this, but I think we should be able get everything working and merged in without it if you prefer.

holopy/core/io/vis.py
```diff
@@ -341,7 +341,7 @@ def show_scatterer_slices(scatterer, spacing):
         voxel spacing for the visualization
     """
     vol = scatterer.voxelate(spacing, 0)
-    show2d(vol)
+    Show2D(vol)
```
barkls, line 344: You should try calling `hp.core.io.vis.display_image` before passing to `Show2D` That function is meant to massage any type of input to a standardized xarray output for visualization or saving. Hopefully that removes the need to check if im has a `voxel` attribute later on. `display_image` is a bit messy, so feel free to ask me if you aren't sure of anything.

barkls:
Hey @anna-wang any progress on this? If you are satisfied just having something that works and would rather pass on extending it any further, that's completely understandable and someone else can work on elaborating on your fix (probably me or @briandleahy). In that case, it would be very helpful if you can provide a minimal example of what was broken before. If you would like to keep working on this, there's no rush, and I'm happy to help with any questions or difficulties that come up!

barkls:
It's possible some overzealous protections were preventing you from pushing to this branch, even though you created it. I've just relaxed those protections.

anna-wang:
@barkls - I'm so sorry it's taken me so long to look into this! I really appreciate your comments though, turns out my messy workaround was totally unnecessary and calling hp.core.io.vis.display_image was all that was needed! Thanks so much for looking into it.

barkls:
I'm glad it ended up being a simple fix and I'm happy that `hp.core.io.vis.display_image` is doing its job! It would still be nice to get a test that captures the failing behaviour. Let me know if you want to write one or if we should assign it to someone else.

anna-wang:
@barkls sorry again for the delay! I know it's simple but I'm a bit swamped, if you could assign to someone else that would be great. This is the code that was failing previously.

```
import holopy as hp
from holopy.scattering.scatterer import Indicators, Sphere, Scatterer
import numpy as np

s1 = Sphere(r = .5, center = (0, 0, 0))
s2 = Sphere(r = .51, center = (0, 0, 0))

def cap(point):
    return(np.logical_and(np.logical_and(point[...,2] > 0, s2.contains(point)),
           np.logical_not(s1.contains(point))))
indicators = Indicators([s1.contains, cap],
                        [[-.51, .51], [-.51, .51], [-.51, .51]])
janus = Scatterer(indicators, (1.34, 2.0), (5, 5, 5))

hp.core.io.vis.show_scatterer_slices(janus,spacing=0.005)
```

I believe this also failed:
```
import holopy as hp
from holopy.scattering.scatterer import Indicators, Sphere, Scatterer
import numpy as np

s = Sphere(r = .5, center = (0, 0, 0))

hp.core.io.vis.show_scatterer_slices(s,spacing=0.005)
```

Thank you!

barkls:
Great, thanks for the examples! I'll look into it...

barkls:
closing in favour of #347

PR #335, opened 2020-04-21
Tighten SuppressOutput and clean up Scatterers

barkls:


barkls:
Hmm meant for this to be a draft.

I want to wait a bit before merging the context manager in case it isn't actually fixed

barkls:
Tests fail...

Looks like Travis is up to some funny business managing output streams:
https://stackoverflow.com/a/58166080

barkls:
I had to go a bit down the Volkswagen route after all. I tried to dig around a bit to figure out what exactly Travis is doing with stdout but it doesn't seem very well documented.

holopy/scattering/scatterer/ellipsoid.py
```diff
@@ -25,19 +25,8 @@
 
 import numpy as np
 
-from .scatterer import CenteredScatterer, Indicators
-from ..errors import InvalidScatterer
-from functools import reduce
-
-def isnumber(x):
```
briandleahy, line 32: Thanks for deleting this unused code from the repo!!

holopy/core/utils.py
```diff
@@ -43,22 +44,33 @@
...
+        except io.UnsupportedOperation:
+            self.stdout_behaves_normally = False  # ie running on travis
+            self.std_out = 1
+        else:
+            self.stdout_behaves_normally = True
+
+    def _redirect_stdout(self, destination_fileno):
+        os.dup2(destination_fileno, self.std_out)
+        if self.stdout_behaves_normally:
+            sys.stdout.close()
```
briandleahy, line 61: Question, not a comment: Do we want to close `sys.stdout` on exiting the context manager?

barkls: Yes, that is intended. It will flush and close the existing `sys.stdout`. The next line sets up a new `sys.stdout` as a low level text buffer. I don't completely understand everything that's going on, but it seems to provide a level of redundancy on top of redirecting output with `os.dup` and `os.dup2`, and stops the rare (approx 1%) issues where output either wasn't being suppressed or wasn't being reinstated. Since both were happening sometimes we need the patch on each end. Unfortunately in an inference calculation something that occurs in 1% of calculations ends up being pretty frequent!

barkls:
Thanks for the review. I think it's ready to merge!

PR #336, opened 2020-04-30
Remove xarray version restriction on travis

briandleahy:


briandleahy:
There are a lot of merge commits here which is a bit overkill; it's probably best to squash-merge this in for the 1 commit.

PR #337, opened 2020-05-04
Scattering doc update

jeromefung:
Added a remark in the User Guide noting that one might want to use Mie rather than Multisphere for multiple widely-separated spheres.

This stemmed from a recent email exchange I had with a HoloPy user who was obtaining some unphysical results when calculating holograms with the default scattering theory (Multisphere) for multiple spheres separated by many particle diameters. There's nothing in the documentation to suggest using Mie in such a case.

The real issue the user was encountering was that Multisphere expands the scattered fields in vector spherical harmonics whose origin is at the center of mass of all the spheres. Apparently, the default limits on the expansions by scsfmo1b weren't enough for the user's situation (but in that situation, optical coupling was negligible).

Based on the introductory level at which the User Guide is written, I didn't try to explain the issue in detail, but my proposed change does suggest using Mie for widely separated spheres. I'm happy to add a more detailed explanation if that would be preferred.


briandleahy:
Awesome, thanks Jerome. 

@barks is in the process of writing a slightly more detailed "user guide" for the scatterers; it might be good to include some discussion of this there, @barkls .

Going ahead and merging this in.

barkls:
Thanks for doing this @jeromefung! It might be nice to have this switch [happen automatically](https://github.com/manoharan-lab/holopy/blob/d7f6f4295e6249462dd242734f52903df9c6a502/holopy/scattering/interface.py#L112). Do you know off-hand what separation distances scsmfo1b will tolerate? Alternatively we could set a threshold for when multiple scattering can be safely ignored.

jeromefung:
The question of what separation distances scsmfo1b tolerates (given default
parameters for the maximum number of orders, etc) is one that I'd have to
investigate in more detail. (We currently suppress the output from
scmsfo1b, as we should, but IIRC it spits out various convergence checks,
and that's probably what we would want to check.)

I'd be more in favor of raising a Warning. My 2011 JQSRT paper works out a
criterion for ignoring multiple scattering that involves interparticle
separations and the particle extinction efficiency. It might not be too
computationally severe to calculate Q_ext for the largest particle,
calculate the criterion for the largest particle separation, and raise a
Warning suggesting the use of Mie if the result is tiny (e.g., < 0.01).

Best,
Jerome

Issue #342, opened 2020-05-14
example data files missing in conda distributions


barkls:
I don't know how widespread this is, but sometimes the conda distribution doesn't properly bring the example data files along for the ride, which means users can't go through the tutorials. It would be nice to get some examples of systems on which this has happened.

barkls:
This problem arises because the exampledata folder has no `__init__.py` so conda doesn't pick it up. This means maybe our exampledata has never worked from a conda distribution? I don't think that's the case so maybe something has changed about what conda/conda-forge includes by default. Here are a few solutions:

1) Add a `__init__.py` file to `hp/core/tests/example_data`. I don't really like this solution since there are no python files there and you shouldn't be able to import this as a submodule. 

2) Adjust our conda recipe on conda-forge to [explicitly include](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#specifying-files-to-include-in-output) all files in the exampledata directory in the package. I think this is the better option.

briandleahy:
If you're looking for input, I agree with you -- this seems to be a conda-specific problem, so a conda-specific solution is probably best (i.e. your number 2 above). We can always come back to this if it ends up being a more general problem.

mdhoney:
> It would be nice to get some examples of systems on which this has happened.

This happened for me.  HoloPy installed fine with conda, but the whole tree below `tests` was missing.  I ended up manually grabbing the files as a ZIP.  What kind of information would be helpful to you?

PR #347, opened 2020-06-01
Develop

barkls:
Supersedes #334 and also cleans up test import structure


barkls:
removing the test init files will prevent sphinx apidoc from building files for the test directories since they won't be treated as submodules any more. Tests pass with them removed, but there might be unintended consequences, such as files not getting carried around by conda. I don't think anyone is trying to run tests of a conda distribution though, and inference tests never had __init__.py to begin with.

barkls:
I don't think we want to have the tests in the HoloPy namespace. Is there a use case for importing test code (including common) when you can't just run the file directly? I see imports as being useful when you want to be able to run the code but not edit it, and I'm not sure when this would ever apply to tests.

That being said, I don't feel strongly about it and we can certainly put the __init__ files back in, which could also have the side-effect of resolving #342. Currently things are inconsistent with some test directories having an `__init__.py` and others not, so we could apply a standard either way.

briandleahy:
Sounds good. We can always change it back if we want later. 

Merging this in now.

briandleahy:
Thanks @barkls . I thought I dismissed them myself earlier.

PR #351, opened 2020-06-25
Fix348

briandleahy:
Allows the `holopy.inference.model.PerfectLensModel` to use an alpha parameter.


briandleahy:
Failing test is an issue with importing numpy.

> Importing the numpy C-extensions failed. This error can happen for
> many reasons, often due to issues with your setup or how NumPy was
> installed.
> We have compiled some common reasons and troubleshooting tips at:
>    https://numpy.org/devdocs/user/troubleshooting-importerror.html
> Please note and check the following:
>  * The Python version is: Python3.6 from "/home/travis/miniconda3/envs/testenv/bin/python"
>  * The NumPy version is: "1.18.5"
> and make sure that they are the versions you expect.
> Please carefully study the documentation linked above for further help.


This seems like a numpy issue and not our problem. Thoughts? Pin Travis to 1.18.4 temporarily?

barkls:
I'll review the PR later. For now, can you verify that the problem is numpy 1.18.4 --> 1.18.5, i.e. that it works in one case and not in the other? I think you don't use conda so it might not be as straightforward to test on your machine. If you test on Travis, you should open a new branch or PR for that purpose so as not to clutter up the commit log here of something that will eventually be merged in.

Sorry, I know this is completely unrelated to what you were trying to accomplish here. I will open a docs-related PR sometime before our next meeting so I could deal with it then as well.

briandleahy:
The failing build is definitely a problem with travis -- my test PR #352 fails the build even though 0 lines are changed. 

However, I'm a bit confused what the problem could be, since it seems that numpy 1.18.5 was [released 22 days ago](https://github.com/numpy/numpy/releases). Checking if pinning to 1.18.4 fixes the problem now.

briandleahy:
Pinning to 1.18.4 fixes the problem on #352 . 

Thoughts @barkls ?

barkls:
There are a few really weird things going on here. If pinning Travis to 1.18.4 fixes the problem then that seems fine for now.

PR #358, opened 2020-07-14
Adjustments to inference interface

### LINKED ISSUE #354: fit and sample functions don't take strategy arguments
Right now the convenient `fit` and `sample` functions always use the default strategy. It is not possible to choose strategy by name through a keyword argument, as described in [the docs](https://holopy.readthedocs.io/en/master/users/fit_tutorial.html#customizing-the-algorithm).

Thanks to @anna-wang for finding this bug.

barkls:
Prompted by #354 (Fixes #354). The docs imposed an arbitrary restriction that `fit` and `sample` high-level functions could only take strategy names as strings, and not actual `Strategy` objects, whereas the `model.fit` and `model.sample` methods could handle either strategy names or objects. I found this confusing and arbitrary so I taught the high-level functions to use objects as well as names. However, this resulted in there being no difference in functionality between the high-level functions and the model methods. I decided to deprecate the model methods so there's only one way to do things, and it's the way that is most flexible and best documented. Note that these methods were introduced as part of the 3.0 development cycle and aren't described very well in the docs so we should be able to remove them with the rest of the old fitting infrastructure.

### DIFF
--- a/holopy/inference/interface.py
+++ b/holopy/inference/interface.py
@@ -22,8 +22,20 @@ import numpy as np
 
 from holopy.scattering import Scatterer
 from holopy.scattering.scatterer import _interpret_parameters
-from holopy.inference.model import ALL_STRATEGIES, Model, AlphaModel
+from holopy.inference.model import Model, AlphaModel
 from holopy.inference.prior import Uniform
+from holopy.inference.nmpfit import NmpfitStrategy
+from holopy.inference.scipyfit import LeastSquaresScipyStrategy
+from holopy.inference.cmaes import CmaStrategy
+from holopy.inference.emcee import EmceeStrategy, TemperedStrategy
+
+DEFAULT_STRATEGY = {'fit': 'nmpfit', 'sample': 'emcee'}
+ALL_STRATEGIES = {'fit': {'nmpfit': NmpfitStrategy,
+                          'scipy lsq': LeastSquaresScipyStrategy,
+                          'cma': CmaStrategy},
+                  'sample': {'emcee': EmceeStrategy,
+                             'subset tempering': TemperedStrategy,
+                             'parallel tempering': NotImplemented}}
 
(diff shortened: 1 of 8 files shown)


holopy/inference/interface.py
```diff
@@ -22,28 +22,57 @@
...
+    if strategy is None:
+        strategy = DEFAULT_STRATEGY[operation]
+    if isinstance(strategy, str):
+        strategy = ALL_STRATEGIES[operation][strategy]
+    if not hasattr(strategy, operation):
+        raise ValueError("Cannot {} with Strategy of type {}.".format(
+            operation, type(strategy).__name__))
+    try:
+        strategy = strategy()
+    except TypeError:
```
briandleahy, line 73: I'm a little worried by this, because there could conceivably be a TypeEror in the strategy. I think an `if isinstance()` call might be better here.

barkls:
Travis failed on a test that had passed on my own machine and shouldn't be related to any code I touched. I restarted the Travis build and it passed, so something strange is going on that I expect was a pre-existing problem with reproducibility in one of the tests. Unfortunately I think there's now no record of the failing build or which test was problematic.

Ready to merge @briandleahy

briandleahy:
Yes, I saw. I believe it was a test on a BoundedGaussian, testing that all samples returned were within the bounds. IIRC there was no seed test in the test, and it was only testing for 10 samples.  (I was planning on discussing it with you at the holopy meeting yesterday, but I forgot that we were not meeting until next week.)

To me this seems like an edge case that happens once in a while (samples are generated outside the bounds).

I would say "someone" should do the following:
* Find exact failing test.
* Raise an issue that this test is finicky.
* Set a seed and increase the number of samples to a few thousand to see if the test fails and therefore what the bug is. 
* Fix the bug.

I can take a stab at that when I get a chance. 

In the mean time, I'll approve this PR.

PR #361, opened 2020-07-28
Fixfailingdoctest

### LINKED ISSUE #360: Failing Doctests
Doctests [fail on Travis](https://travis-ci.com/github/manoharan-lab/holopy/builds/177471598) after #357. This was obscured before because the unittests were failing on that PR so doctests never ran. 

The problem looks pretty simple - some documentation including example code was moved from scattering tutorial to scattering user guide and it looks like some import statements didn't follow.

briandleahy:
Fixes #360


barkls:
Thanks for this!

PR #364, opened 2020-08-14
Fix #359 and #362

### LINKED ISSUE #359: Rarely failing test - bug with BoundedGaussian Sampling?
The test `hp.inference.tests.test_prior.TestBoundedGaussian.test_sample` failed at least once on Travis in the context of #358. 

Here is @briandleahy's plan of attack pasted from a comment in that thread:

> 
> IIRC there was no seed test in the test, and it was only testing for 10 samples.
> 
> To me this seems like an edge case that happens once in a while (samples are generated outside the bounds).
> 
> I would say "someone" should do the following:
> 
> Find exact failing test. [**done**]
> Raise an issue that this test is finicky. [**done**]
> Set a seed and increase the number of samples to a few thousand to see if the test fails and therefore what the bug is.
> Fix the bug.

### LINKED ISSUE #362: missing numexpr warning
There's a warning about numexpr being missing and Lens calculations being slow. Right now this appears as soon as you import holopy. It should be changed to appear when you use it - probably when you call the LensTheory `__init__` - to match the other optional dependencies in HoloPy - emcee, cma, Fortran, DDA, schwimmbad, etc.

briandleahy:
Three small commits that fixes #359, fixes #362, and tags a slow test as slow.

### DIFF
--- a/holopy/inference/tests/test_interface.py
+++ b/holopy/inference/tests/test_interface.py
@@ -44,7 +44,7 @@ class TestUserFacingFunctions(unittest.TestCase):
     def test_cannot_sample_without_model(self):
         self.assertRaises(ValueError, sample, DATA, Sphere())
 
-    @attr('fast')
+    @attr('slow')
     def test_sample_function_calls_model_sample(self):
         result = sample(DATA, SimpleModel())
         self.assertTrue(isinstance(result, SamplingResult))
--- a/holopy/inference/tests/test_interface.py
+++ b/holopy/inference/tests/test_interface.py
@@ -44,11 +44,11 @@ class TestUserFacingFunctions(unittest.TestCase):
     def test_cannot_sample_without_model(self):
         self.assertRaises(ValueError, sample, DATA, Sphere())
 
-    @attr('slow')
+    @attr('fast')
     def test_sample_function_calls_model_sample(self):
-        result = sample(DATA, SimpleModel())
+        strategy = EmceeStrategy(nsamples=1)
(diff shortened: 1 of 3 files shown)


holopy/inference/tests/test_interface.py
```diff
@@ -44,7 +44,7 @@ class TestUserFacingFunctions(unittest.TestCase):
     def test_cannot_sample_without_model(self):
         self.assertRaises(ValueError, sample, DATA, Sphere())
 
-    @attr('fast')
+    @attr('slow')
```
briandleahy, line 47: @barkls, is there an easy way to make this slow test run faster?

barkls: Yes. We aren't actually trying to test HoloPy's interactions with emcee here, just that the handy interface works correctly. This test is really doing 2 things and they should be broken up.

1) Check that the high-level `sample` function calls the `strategy.sample` method. This can be made fast by defining a dummy test strategy with a `sample` method that doesn't actually sample with emcee.

2) Check that the high-level `sample` function uses the right strategy if none is provided. This is actually redundant with another test further down in the file (`StrategyHandling.test_default_sampling_strategy_is_emcee` - should really be renamed `TestStrategyHandling`) so we can remove it here but it's probably just as slow there. One way to speed this operation up in the other test might be to design the sampling to fail with a pathological data or model. We could then check an `assertRaises` from within the `EmceeStrategy.sample` method to make sure it directs to the right place, but doesn't get as far as actually sampling.

Thoughts? Let me know if anything I wrote is unclear...

PR #365, opened 2020-08-26
Refactor parameters

### LINKED ISSUE #256: can't save Model objects with xarray properties

### LINKED ISSUE #353: Can't fit multiple particles
Right now you can't use the convenient `fit` function on `Scatterers` objects. 

This is because `hp.inference.interface.parameterize_scatterer` calls `hp.scattering.scatterer.scatterer._interpret_parameters`, which was changed in #279 to colons in parameter names as special characters to denote dictionaries or xarrays of parameter values. Unfortunately `hp.scattering.scatterer.Scatterers.parameters` also uses colons to denote different particles within a `Scatterers` object. This hadn't been noticed as a problem because `_interpret_parameters` is mostly used in `from_parameters`, and `Scatterers` defines their own.

barkls:
Fixes #353. Fixes #256.

Moved all parameter handling from Scatterer class to Model class. This fixed some problems with tied parameters and model flexibility. The Model now stores scatterer, alpha, optics, and hopefully theory in the form of a "map" that tells it how to reconstruct the object with appropriate parameter substitutions. This PR largely rewrites the model.py file and that is the most important place to look to understand what's going on. Everything else is mostly just adapting to those changes.

There are a few scatterer API changes that I want to mention:
1) scatterer.parameters used to give a flat dictionary, e.g. `{n.real: Uniform(1.58, 1.60), n.imag: 0.001, 'center.0': 10, 'center.1':10, 'center.2':Uniform(0, 100)}`. After this PR it matches the initial arguments to call the scatterer: `{'n': ComplexPrior(Uniform(1.58, 1.60), 0.001), 'center': (10, 10, Uniform(0, 100))}` .
2) `scatterer.from_parameters` used to always give a definite value - it would substitute all priors with their best guess. With this PR it just returns priors without modification.
3) Previously you could call `calc_holo` on a parameterized scatterer and it would similarly substitute priors with their best guess. This PR requires you to define a model object if you want to do that (but you were going to define a model anyway if using a parameterized scatterer).

There's also a big model API change that I'm less sure about and would appreciate some feedback on:
Previously, model methods (lnprior`, lnlike, residuals, lnposterior, forward, and the new scatterer_from_parameters) wanted a dictionary of parameter values to evaluate at, e.g. {'r': 0.5, 'n': 1.59, 'alpha': 0.8}. With the changes in this PR, I found a list of parameter values was more natural, and there was a lot of switching back and forth between them if we wanted to keep dictionaries everywhere. After this PR, the methods expect a list of parameter values in a specific order e.g. [0.5, 1.59, 0.8]. They can still take a dictionary for backwards compatibility but it is deprecated.

Unfortunately both formats have advantages, and this is where I'd like input. The dictionary is more human readable and lets you easily test how different parameter values would affect the hologram, for example. It matches `model.parameters` which is the main user-facing way to see what the model parameters are. A list makes for cleaner code, and a simpler interface for inference algorithms, both within holopy an externally. This corresponds to `model._parameters`, which does all the work but isn't user facing.


briandleahy:
It looks like some of the tests are failing. Do you want me to take a look at it now, or wait until you've addressed the tests?

barkls:
Failing tests should be trivial to fix. Review now will be helpful since I expect there might be a few rounds of changes. Thanks!

holopy/core/holopy_object.py
```diff
@@ -107,9 +107,4 @@ def __eq__(self, other):
             # Note, this is possibly a slightly weak form of equality, but well behaved holopy objects are equal if their _dict's are equal
             return self._dict == other._dict
         else:
-            return False
-
-    def like_me(self, filter_none=True, **kwargs):
-        if filter_none:
-            kwargs={key: val for key, val in kwargs.items() if val is not None}
-        return self.__class__(**dict(self._dict, **kwargs))
+            return False
```
briandleahy, line 110: Thanks for deleting code that wasn't being used!

holopy/core/metadata.py
```diff
@@ -453,11 +453,6 @@ def dict_to_array(schema, inval):
         msg = ("Dictionary could not be converted to DataArray because " +
                "reference grid has no dimensions with matching coords")
         raise ValueError()
```
briandleahy, line 455: again, thanks for deleting unused code!

holopy/inference/interface.py
```diff
@@ -74,36 +74,59 @@ def validate_strategy(strategy, operation):
     return strategy
 
 
-def make_default_model(base_scatterer, fitting_parameters):
+def make_default_model(base_scatterer, fitting_parameters=None):
     if fitting_parameters is None:
-        fitting_parameters = base_scatterer.parameters.keys()
+        fitting_parameters = list(base_scatterer.parameters.keys())
```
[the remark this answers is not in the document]
briandleahy: Did you find this because a test failed? If so, it might be a sign that we need to test this better. (Unless if the parameters don't actually need to be a list.)

holopy/inference/interface.py
```diff
@@ -74,36 +74,59 @@ def validate_strategy(strategy, operation):
...
+
+def replace_center(parameters, key):
+    index = COORD_KEYS.index(key[-1])
+    prefix = key[:-1]
+    try:
+        val = parameters.pop(key)
+    except KeyError:
+        pass
+    else:
+        parameters[prefix + 'center'][index] = val
```
briandleahy, line 109: I'm a little confused as to what this code does, esp. since you said in the PR that
> scatterer.parameters used to give a flat dictionary, e.g. {n.real: Uniform(1.58, 1.60), n.imag: 0.001, 'center.0': 10, 'center.1':10, 'center.2':Uniform(0, 100)}. After this PR it matches the initial arguments to call the scatterer: {'n': ComplexPrior(Uniform(1.58, 1.60), 0.001), 'center': (10, 10, Uniform(0, 100))} .

Can you let me know what I'm missing, either in the comments or when we meet?

holopy/inference/interface.py
```diff
@@ -74,36 +74,59 @@ def validate_strategy(strategy, operation):
...
+    minval = 0 if suffix in ['n', 'r'] else -np.inf
+    if isinstance(guess_value, (list, tuple, np.ndarray)):
+        if suffix == 'center':
+            subkeys = COORD_KEYS
+        else:
+            subkeys = [suffix + '.' + str(i) for i in range(len(guess_value))]
+        return [Uniform(minval, np.inf, guess_val, prefix + subkey)
+                for guess_val, subkey in zip(guess_value, subkeys)]
+    else:
+        return Uniform(minval, np.inf, guess_value, key)
```
briandleahy, line 132: Same as above, here. (I'm a little confused as to what this is doing, since it seems to assume keys have a colon in them, when my impression of the new version was that keys had no colon.

It might be a little clearer if you split out the code that gets the guess value?? Or maybe I'm missing something completely.

holopy/inference/nmpfit.py
```diff
@@ -165,7 +166,7 @@ def minimize(self, parameters, obj_func):
 
         def resid_wrapper(p, fjac=None):
```
briandleahy, line 167: I know this is from before your commit, but it would be good if we could change the single-character variable name to something more descriptive.

holopy/inference/tests/common.py
```diff
@@ -25,19 +25,20 @@
...
+        self._maps = {'optics': {}}
 
     def _residuals(self, pars, data, noise):
         return self.lnposterior(pars, data, None)
 
     def lnposterior(self, par_vals, data, dummy):
         x = par_vals
         data = np.array(data)
-        return -((x[self._parameters[-1].name] - data)**2).sum()
+        return -((x[-1] - data)**2).sum()
```
briandleahy, line 41: Just a comment so we can talk about this in our next meeting -- if the lnposterior for SimpleModel is so simple, why is the sampling so slow?

barkls:
tests should pass now. Sorry to push while you're mid-review!

holopy/inference/tests/test_interface.py
```diff
@@ -185,14 +184,45 @@ def test_make_default_model_construction(self):
...
-        self.assertEqual(parameters_list, ['center.0', 'p', 'center.1'])
+    def test_parameterize_scatterer_xyz(self):
+        scatterer = parameterize_scatterer(Sphere(center=[0, 0, 0]), ['x'])
+        self.assertTrue(isinstance(scatterer.center[0], prior.Prior))
+
+    @attr('fast')
+    def test_parameterize_scatterer_center(self):
+        scatterer = parameterize_scatterer(Sphere(center=[0, 0, 0]), ['center'])
+        expected = prior.Uniform(-np.inf, np.inf, 0, 'x')
+        self.assertEqual(scatterer.center[0], expected)
```
briandleahy, line 199: Does `parameterize_scatterer(Sphere(center=[0, 0, 0]), ['center'])` return all 3 priors for x, y, and z, or just x? If so, we could do
```python
for i, xyz in enumerate('xyz'):
    expected = prior.Uniform(-np.inf, np.inf, 0, xyz)
    self.assertEqual(scatterer.center[i], expected)
```

holopy/inference/interface.py
briandleahy: Oh... is this because of `Spheres` vs `Sphere`?

holopy/scattering/interface.py
```diff
@@ -195,17 +193,8 @@ def calc_holo(detector, scatterer, medium_index=None, illum_wavelen=None,
     theory = interpret_theory(scatterer, theory)
     uschema = prep_schema(
         detector, medium_index, illum_wavelen, illum_polarization)
-
```
briandleahy, line 198: ZOMG, did you eliminate the need for this ugly massaging of scaling with color channels??!!!?! If so, :100:

holopy/scattering/scatterer/scatterer.py
```diff
@@ -114,12 +121,9 @@ def parameters(self):
             passed to Scatterer.from_parameters to make a copy of this
             scatterer
         """
-        # classes that have anything complicated happening with their variables
-        # should override this, but for simple classes the variable self._dict
-        # is the correct answer
-        return dict(_expand_parameters(self._dict.items()))
+        return deepcopy(self._parameters)
```
briandleahy, line 124: Why the deepcopy? What gets modified in place if you do a regular copy?

holopy/scattering/theory/scatteringtheory.py
```diff
@@ -277,3 +278,17 @@ def _transform_to_desired_coordinates(cls, detector, origin, wavevec=1):
             cls.desired_coordinate_system)
         return method(original_coordinate_values)
 
+
+def _select_scatterer(scatterer, illum):
```
briandleahy, line 282: Since this is just for multi-wavelength, can we rename this to `_select_multiwavelength_scatterer` or something?

Also, what was happening before -- was there a bug that you caught or is this a new need?

holopy/inference/model.py
```diff
@@ -28,9 +28,79 @@
...
+
+
+def make_complex(real, imag):
+    if isinstance(real, Prior) or isinstance(imag, Prior):
+        return ComplexPrior(real, imag)
+    else:
+        return complex(real, imag)
+
+
+def read_map(map_entry, parameter_values):
```
briandleahy, line 56: Are the `map_entry`s seen by the user? If not, it might be worth making a small class / namedtuple that captures this. You could then subsume some of this more complicated logic into that class, which might reduce the amount of edge cases that can happen when an incorrect argument is passed to as `map_entry`.

briandleahy: The advantage would be that errors would be raised at the correct point. For instance, if you call `map_entry.read_map(parameter_values)`, an error will be raised if `map_entry` is a list, whereas currently things will propagate down the chain and make it harder to debug.

briandleahy: It would also have an advantage of making a `map` a concrete object, which would allow for an easier understanding of how the code works. Right now it seems that the mapping functionality is diffused across multiple regions of the file; making it a helper class would concentrate it in one place.

briandleahy (commented):
Hey @barkls , 

Thanks for doing this. This was clearly a lot of work. 

In general I think things look good. I'm still a bit confused about how the map actually works. Can you explain it to me when we meet next?

I think the three API changes that you bulleted are good. Regarding the dictionary vs the list -- I think this is OK, as long as there is a clear correct order for the parameters to get passed in. One advantage of a dictionary is that the order does not matter. An alternative is to take a dict as inputs and cast to a list right away for the actual inner workings, although I'm not sure if that would be viable.  It would help if you showed me in the code where this is, when we talk about it.

Regarding changes -- I left a few general comments throughout; feel free to take them or leave them. The pep8 bot also made some comments; it would be nice if you could address some of them. (I appreciate you waiting until I took a look at this first though, thanks!)

When we meet next, we can chat about this and make sure I'm understanding things correctly. Then I'll approve and merge. 

Thanks again @barkls !

holopy/scattering/interface.py
briandleahy: It is now taken care of by `scatteringtheory.py` and other places. :100:

holopy/inference/interface.py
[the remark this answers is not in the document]
barkls: Turned out the list casting wasn't actually necessary here, so I've reverted the change.

holopy/inference/interface.py
barkls: Yeah this is to handle the case of Scatterers objects, which might have '2:y' as a user-specified fitting parameter that needs to replaced with '2:center'[1] since that's how it would appear in Spheres.parameters.

holopy/inference/interface.py
barkls: Colons are still used by composite scatterers to denote their individual components.

Automated detection of fitting parameters from scatterers got complicated when scatterers forgot how to interpret themselves. Specifically, the combination of Spheres objects and trying to fit one or more 'center' coordinates ended up producing a bunch of cases that need to be handled. This is why I had suggested not doing the automatic detection for composite scatterers. 

Note that all of this parameter handling in interface.py is just to come up with a Scatterer object containing prior objects for the case where a user calls hp.fit(data, scatterer). More advanced users will define their own Scatterer with prior objects and a Model.

holopy/inference/nmpfit.py
barkls: renamed to parameters

holopy/inference/tests/common.py
barkls: see email thread...

holopy/scattering/theory/scatteringtheory.py
barkls: renamed to `select_scatterer_by_illumination`. In theory it should work for e.g. cross-polarized light sources as well as different wavelengths. 

This was previously handled by the Scatterer itself, but has been moved here as part of simplifying the Scatterer class.

holopy/inference/model.py
barkls: I tried making a class to handle this but it didn't help as much as I hoped due to the following:
1) Some of the map interpretation code couldn't be  easily incorporated into the new class since it was needed to read raw maps (not part of an object) during Model serialization/unserialization.
2) There ended up being a lot of calling back and forth between the model and the new object and I needed to add a whole bunch of new methods/properties to the model class just to retrieve things from the new object while maintaining existing namespaces.
3) It became more convoluted to add a new type of Model, since the forward methods would need to interact with the new class.

All those difficulties aside, it did nicely store all the data currently associated with the Model class, and the Model became a series of wrapper methods specifically for doing inference. This matches nicely with something I remember reading in Clean Code about having classes to store data and classes to do calculations, but ideally not both in the same object.

I agree this would be a good change to make, but I couldn't figure out an elegant way to do it, so I abandoned my efforts. I'm happy to revisit it if anyone thinks the current state of model.py is too messy and/or have suggestions about resolving the above issues.

barkls:
I've applied @briandleahy suggestions here, and made a clearer distinction among Model methods between those that are user-facing (cast parameters dict to list if necessary), and those that are used internally (only accept parameters list).

I didn't end up making a separate class to hold the model parameters and map - it ended up being too tightly integrated into the existing Model class for an easy separation (see a comment thread in the review of this PR for a more detailed description). I can take another stab at this though if others feel strongly about it!

briandleahy (approved):
Looks good to me! I'm happy to merge it in if you're ready. 

With regards to a separate map entry class -- I think this is still a good goal, but we can wait to do it until later. The best is the enemy of the good :smile:

I'll merge this in at the end of the day today, unless you let me know you want me to wait.

barkls:
Hold off on merging. I've discovered a bug with loading old Scatterers
objects since they no longer have ties.

barkls:
Ok I fixed the backwards compatibility bug.

briandleahy (approved):

PR #370, opened 2020-09-25
adda on gpu and improved dpl

golovart:
Adding an option to run ADDA on GPU (adda_ocl) and to select the GPU index.

`dpl`: As mentioned in ADDA manual the dipoles need to be "small compared to both any structural length in the scatterer and the wavelength" (_Section 4. Applicability of the DDA_). For scatterers much smaller than the wavelength they suggest "to use at least 10 dipoles along the smallest dimension", which I tried to implement using the scatterer.bounds.
This is not the perfect solution for complex structures, but in that case user should better specify the dipole size himself as `max_dpl_size`.
Maybe some kind of warning to keep an eye on the dipole size for small scatterers should be added in the documentation?

### DIFF
--- a/holopy/scattering/theory/dda.py
+++ b/holopy/scattering/theory/dda.py
@@ -76,8 +76,8 @@ class DDA(ScatteringTheory):
     it will need excessive memory or computation time for particularly
     large scatterers.
     """
-    def __init__(self, n_cpu = 1, use_gpu = False, gpu_id = None, max_dpl_size=None, use_indicators=True,
-                 keep_raw_calculations=False, addacmd=[],
+    def __init__(self, n_cpu=1, use_gpu=False, gpu_id=None, max_dpl_size=None,
+                 use_indicators=True, keep_raw_calculations=False, addacmd=[],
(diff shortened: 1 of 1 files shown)


barkls:
Thanks for this PR! I just have a couple minor points before merging in:

1) Have you confirmed that this works both with and without a gpu_id specified?

2) If someone sets use_gpu=True, then the `n_cpu` parameter is ignored. It would be nice to raise a warning or exception if n_cpu is also set to some other value to prevent confusion. However, it's not a big deal so I can merge without this check if you  prefer.

golovart:
1. There was a small typo when passing the "gpu_id", now I confirm that it is running on that gpu, which index is passed.
Without specifying it adda chooses the default gpu itself.
2. I added the multi-cpu warning.

Should I correct those warnings above (like too long lines)?

barkls (approved):

barkls:
Looks good! Thanks for the update to dipole scaling as well. No need to worry about the overly long lines. Thanks again for the contribution!

PR #372, opened 2020-10-21
Transformed priors

barkls:
Documentation on using these new things is lacking. We should have a user guides on Model & Prior objects, and also a tutorial using the new features. I'm happy to get this merged in first though

### DIFF
--- a/holopy/inference/model.py
+++ b/holopy/inference/model.py
@@ -30,8 +30,8 @@ from holopy.scattering.errors import (MultisphereFailure, TmatrixFailure,
                                       InvalidScatterer, MissingParameter)
 from holopy.scattering.interface import calc_holo
 from holopy.scattering.theory import MieLens
-from holopy.inference.prior import (Prior, Uniform, ComplexPrior,
-                                    TransformedPrior, generate_guess)
+from holopy.inference.prior import (Prior, Uniform, TransformedPrior,
+                                    generate_guess)
 
 
 OPTICS_KEYS = ['medium_index', 'illum_wavelen',
@@ -65,13 +65,6 @@ def make_xarray(dim_name, keys, values):
         return xr.concat(values, dim=new_dim)
     else:
         return xr.DataArray(np.array(values), coords=[keys], dims=dim_name)
-
-
-def make_complex(real, imag):
-    if isinstance(real, Prior) or isinstance(imag, Prior):
-        return ComplexPrior(real, imag)
-    else:
-        return complex(real, imag)
 
 
 def transformed_prior(transformation, base_priors):
@@ -160,8 +153,6 @@ class Model(HoloPyObject):
             mapped = self._map_dictionary(parameter, name)
         elif isinstance(parameter, xr.DataArray):
             mapped = self._map_xarray(parameter, name)
-        elif isinstance(parameter, ComplexPrior):
-            mapped = self._map_complex(parameter, name)
         elif isinstance(parameter, TransformedPrior):
             mapped = self._map_transformed_prior(parameter, name)
         elif isinstance(parameter, Prior):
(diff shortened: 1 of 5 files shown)


barkls:
oh no! I forgot to update the changelog. Don't merge until I do...

holopy/inference/model.py
```diff
@@ -66,11 +67,11 @@ def make_xarray(dim_name, keys, values):
         return xr.DataArray(np.array(values), coords=[keys], dims=dim_name)
 
 
-def make_complex(real, imag):
```
briandleahy, line 69: What happened to `make_complex`? Is the idea that now we say `n=prior.Gaussian(1.59, 0.1) + 1j * prior.Uniform(0, 1e-3)`?

briandleahy: (I'm mostly asking about documentation.)

holopy/inference/prior.py
```diff
@@ -77,6 +86,25 @@ def __rmul__(self, value):
     def __truediv__(self, value):
         return self * (1/value)
 
+    def __rtruediv__(self, value):
+        return value * TransformedPrior(_reciprocal, self)
```
briandleahy, line 90: Just curious -- why are these not `return TransformedPrior(operator.truediv, [self, value])` and `return TransformedPrior(operator.truediv, [value, self])`?

holopy/inference/prior.py
```diff
@@ -295,21 +289,66 @@ def sample(self, size=None):
...
+                      for bp in self.base_prior]
+        return self.transformation(*guess_vals)
+
+    @property
+    def map_keys(self):
+        return enumerate(self.base_prior)
 
 
-class ComplexPrior(Prior):
+class ComplexPrior(TransformedPrior):
```
briandleahy, line 351: Nice!! :+1:

holopy/inference/prior.py
```diff
@@ -17,20 +17,26 @@
...
 
 EPS = 1e-6
 
 
+def _reciprocal(val):
+    return 1 / val
+
+
 class Prior(HoloPyObject):
     """Base class for Bayesian priors in holopy."""
```
briandleahy, line 41: It might be nice to add a list of methods that `Prior` subclasses are expected implement, for the future devs who may extend these. I think either adding a list in the `Prior` docstring or adding empty methods, like
```python
    def lnprob(self, p)
        raise NotImplementedError("implement in subclass")
```
works. This might be out of the scope of this PR though.

briandleahy (approved):
Looks good to me @barkls . I left a few minor comments, but I don't think they are necessary for this PR. 

At some point we do need to update the docs for this, but that should probably be a separate PR.

Let me know when you want me to merge this in (since you said you wanted to update the change log).

holopy/inference/model.py
barkls: `make_complex` was just a dispatch function used by a Model to unpack a map (e.g. into a scatterer). It became redundant when `ComplexPrior` was rolled into `TransformedPrior` which has its own similar function.

holopy/inference/prior.py
barkls: They probably could be - previously the math behaviour had to be written manually so I tried to use identities to get everything onto `__mul__` and `__add__`. That's less of an issue now, but there is still specific handling for special values like 0 or 1. Do you think it would be more clear to use `operator` functions throughout?

holopy/inference/prior.py
barkls: Nope, the most important `TransformedPrior` methods (`guess` and `sample`) are inherited directly.

I do override some placeholders (`prob`, `lnprob`, `map_keys`) in `TransformedPrior` with more intelligent behaviour specific to the `ComplexPrior`. I do think there's a future in which other types of transformations get their own classes (e.g. arctan or other trig), and they would need to override these defaults as well. Let me know if there's a better way to do this.

Finally, there are some properties (`real` and `imag`) that are just convenience ways of accessing `base_prior` components that also maintain backwards compatibility. Actually `map_keys` was added to maintain backwards compatibility as well. I initially had `Model` just follow the `TransformedPrior` behaviour without asking the object, but that didn't match previous parameter names for `ComplexPrior` objects.

holopy/inference/prior.py
barkls: I'm planning to write a user guide for the `Prior` class at some point, like the ones for `Scatterer` and `ScatteringTheory`. Would that be sufficient or do you think I should put something in the code as well?

holopy/inference/prior.py
briandleahy: Personally, I think that keeping the same pattern throughout would be the clearest, but I'm not sure. For instance, if you are end up doing a lot of error handling, it could be cleaner to call the `__mul__` like you are doing here. 

It's up to you.

holopy/inference/prior.py
briandleahy: :+1:

holopy/inference/prior.py
briandleahy: I think something in the code as well would be good. (Just because someone could update the code in the future and forget to update the user guide.) Again, not super urgent though.

barkls:
Thanks for the review @briandleahy! Ready to merge (assuming tests pass)

PR #378, opened 2020-12-10
Composescatteringtheory v2

briandleahy:
The same as #377, but modified to keep the git blame for each line of code despite the split of `scatteringtheory` into two files. I've kept the commit hashes exactly the same.

Since this PR should be the one that gets merged in, I've copied the notes from #377 below.

This PR composes out a lot of the work of the old `ScatteringTheory` class into two new classes, a (much lighter) `ScatteringTheory` and a new `ImageFormation` class.

**Rationale**
Most of the old `ScatteringTheory` code was actually for calculating images and not for scattering calculations per se. This made creating and modifying new scattering theories unwieldy. In particular, it was difficult to know which methods needed to be implemented to create a new scattering theory. This is now fixed. Moreover, while most of the time in a hologram calculation (say) was spent in the old `ScatteringTheory` class, that time is mostly in dealing with xarrays and not in doing scattering calculations per se. Now that these two responsibilities are separated, performance optimizations will be easier to implement.

**Code Changes**
The code that was previously in `holopy.scattering.theory.scatteringtheory` is now split between `holopy.scattering.theory.scatteringtheory` and `holopy.scattering.imageformation`. In addition, there were some minor changes to the individual scattering theories, particularly the `Lens` theory, to work with the new changes. (The `Lens` theory relied on methods which were moved to the `ImageFormation` class; now it only relies on `ScatteringTheory` methods.)

**API Changes**
All of these changes are internal and not visible to outside users.


briandleahy:
Arg, this preserves the blame for `imageformation.py` but not `scatteringtheory.py` Not sure what to do here...

barkls:
Here's a series of blog posts about how to do this. It looks from the first one like you might be able to rename *both* files before merging, and then rename one back afterwards, although he doesn't explicitly say that. The second and third methods are a bit more convoluted but he specifically shows they do work.
Sorry if I misled you!

https://devblogs.microsoft.com/oldnewthing/20190916-00/?p=102892
https://devblogs.microsoft.com/oldnewthing/20190917-00/?p=102894
https://devblogs.microsoft.com/oldnewthing/20190918-00/?p=102901

briandleahy:
Superseded by #379

PR #379, opened 2020-12-11
Composescatteringtheory keepblame

briandleahy:
The same as #377, but modified to keep the git blame for each line of code despite the split of `scatteringtheory` into two files. I've kept the commit hashes exactly the same.

Since this PR should be the one that gets merged in, I've copied the notes from #377 below.

This PR composes out a lot of the work of the old `ScatteringTheory` class into two new classes, a (much lighter) `ScatteringTheory` and a new `ImageFormation` class.

**Rationale**
Most of the old `ScatteringTheory` code was actually for calculating images and not for scattering calculations per se. This made creating and modifying new scattering theories unwieldy. In particular, it was difficult to know which methods needed to be implemented to create a new scattering theory. This is now fixed. Moreover, while most of the time in a hologram calculation (say) was spent in the old `ScatteringTheory` class, that time is mostly in dealing with xarrays and not in doing scattering calculations per se. Now that these two responsibilities are separated, performance optimizations will be easier to implement.

**Code Changes**
The code that was previously in `holopy.scattering.theory.scatteringtheory` is now split between `holopy.scattering.theory.scatteringtheory` and `holopy.scattering.imageformation`. In addition, there were some minor changes to the individual scattering theories, particularly the `Lens` theory, to work with the new changes. (The `Lens` theory relied on methods which were moved to the `ImageFormation` class; now it only relies on `ScatteringTheory` methods.)


briandleahy:
`git blame` is correct on this PR. Closing the other two.

briandleahy:
There is also no code difference between this and #377, as evidenced by a `git diff 7cf2e484 21287449`

barkls:
Nice work getting the blame history to follow!

I hate to be a pain, but can you update the releasenotes file even though the changes aren't user-facing? Sorry I didn't think of it earlier...

Also see my minor note on #377.

briandleahy:
The release notes should be updated in a5c62fb9

barkls (approved):

PR #385, opened 2021-02-16
Inference docs

### LINKED ISSUE #299: Travis pins version numbers
As of 3.3 release, travis.yml pins a few things that should be updated in future.

I've edited this comment to keep the below list up to date:
python 3.6 (for compatibility with h5netcdf 0.7.4) - [code runs with DeprecationWarnings]
h5py 2.9 (for compatibility with h5netcdf 0.7.4) - [code runs with DeprecationWarnings]
yaml < 0.2 or pyyaml < 5.2 - [code doesn't run]
scipy < 1.4 [code doesn't run]

### LINKED ISSUE #376: docs links in readme are out of date
It might be better to just have a single link to all the documentation to prevent stale links in future.

barkls:
Main thing is adding TransformedPrior to tutorial, viewable on [readthedocs](https://holopy.readthedocs.io/en/inference_docs/tutorial/fit_tutorial.html#transforming-priors). Lots of smaller changes as well - see commits/changelist.

Fixes #299. Fixes #376.


barkls:
I don't feel the need to make those particular PEP8 changes, but I can if you think it's a good idea @briandleahy.

barkls:
I stopped raising exceptions on `FutureWarning` to get travis to pass. I think this is ok because what we really want to make sure we catch is `DeprecationWarning`. When a package bumps versions and uses `FutureWarning`, other packages we have as dependecies won't update right away, e.g. here `h5py` updated, but `h5netcdf` is calling it in a way that generates the `FutureWarning`. We could pin `h5py` until `h5netcdf` updates, but I think it's better to just ignore the `FutureWarning`.

README.rst
```diff
@@ -1,8 +1,9 @@
 Holography and Light Scattering in Python
 =========================================
 
-.. image:: https://travis-ci.org/manoharan-lab/holopy.png
-    :target: http://travis-ci.org/manoharan-lab/holopy
+.. image:: https://travis-ci.com/manoharan-lab/holopy.svg?branch=develop
```
briandleahy, line 4: Good catch.

holopy/inference/interface.py
```diff
@@ -78,7 +78,7 @@ def make_default_model(base_scatterer, fitting_parameters=None):
     if fitting_parameters is None:
         fitting_parameters = base_scatterer.parameters.keys()
     scatterer = parameterize_scatterer(base_scatterer, fitting_parameters)
-    alpha_prior = Uniform(0, 1, guess=0.7, name='alpha')
+    alpha_prior = Uniform(0.5, 1, name='alpha')
```
briandleahy, line 81: What is the logic for this change here?

barkls: fits were going to alpha -> 0 if the initial guess wasn't great. I'd prefer a lower bound with more justification if you have one...

holopy/inference/third_party/nmpfit.py
```diff
@@ -1399,15 +1399,15 @@ def iterfunct(myfunct, p, iter, fnorm, functkw=None,
 
                 ## Fill in actual covariance matrix, accounting for fixed
                 ## parameters.
-                self.covar = numpy.zeros([nn, nn], numpy.float)
+                self.covar = numpy.zeros([nn, nn], float)
```
briandleahy, line 1402: Just out of curiosity, why did you delete the explicit `numpy` module name? Was it causing a bug?

barkls: deprecated as of numpy 1.20. Apparently it just points to build in python float anyway.

barkls: `np.float64` is still a thing, but `np.float` was just another name for `float` and they're killing it. Same with `int`.

holopy/inference/interface.py
briandleahy: I'm not sure there is any good reason for anything regarding alpha. But I would say (my 2 cents):
* Keep `guess = 0.7`, since that is typically what we find is a good fit and since that does provide some information to external users as to what to expect for alpha.
* The alpha=0 is a problem, but perhaps we can just set the lower bound a little higher? 0.5 is just a little too close to "typical" values in my opinion, although maybe that is a good thing. 

Just my 2 cents. I'm fine with whatever you decide, now that I know the change was deliberate :)

briandleahy (approved):
Hi @barkls , 

Everything looks good to me, aside from the detail on the failing doctest. 

When you let me know that you're ready to merge, and the doctest passes, then I'll merge it in. 

Thanks for doing this!

holopy/inference/third_party/nmpfit.py
briandleahy: (I'm just curious, no need to change anything.)

holopy/inference/interface.py
barkls: This is in the user-friendly functional interface, so the goal should be to do whatever will just "work" - I think keeping alpha in a typical range is a good thing. These users don't know anything about alpha. I had earlier suggested fixing it alpha at 1.0 for these calculations, but others rightfully pointed out this isn't what tends to work best. I could maybe fix it at 0.7 though or leave it as a uniform prior over [0.5, 1]. This has a guess at 0.75, which I thought was also reasonable, but I can explicitly set it to 0.7 if you think that's better.

barkls:
Thanks for the review @briandleahy! Tests passing now. I'm happy with the `hp.fit` alpha where it is now (uniform over [0.5, 1], no explicit guess given), but I'm also happy to change it if you prefer.

holopy/inference/interface.py
briandleahy: Ah I forgot that the guess defaults to the midpoint of the interval. This is fine for me!

PR #386, opened 2021-02-19
Mielensab

briandleahy:
This extends `holopy` with the ability to calculate the image of single spheres through a lens with spherical aberration of arbitrary order, at no extra computational cost compared to that for an unaberrated lens (and therefore still faster than the calculations without a lens).

@carolinesmartin has been using this for about 1 year now, so the code has been pretty well stress-tested. 

@barkls let me know if this should be added to a user guide.


holopy/scattering/theory/mielens.py
```diff
@@ -128,3 +131,33 @@ def raw_fields(self, positions, scatterer, medium_wavevec, medium_index,
...
+        return field_calculator
+
+
+class AberratedMieLens(MieLens):
+    def __init__(self, spherical_aberration=0.0, lens_angle=1.0,
+                 calculator_accuracy_kwargs={}):
+        super(AberratedMieLens, self).__init__()
+        self.lens_angle = lens_angle
+        self.spherical_aberration = spherical_aberration
+        self.calculator_accuracy_kwargs = calculator_accuracy_kwargs
```
barkls, line 151: Two of these lines (`lens_angle` and `calculator_accuracy_kwargs`) are sort of redundant with the call to `super.__init__()`, except that you don't pass them through to the `MieLens` init. Is there a specific reason you chose to do things this way?

holopy/scattering/theory/mielensfunctions.py
```diff
@@ -264,11 +264,60 @@ def _interpolate_and_eval_mielens_i_n(self, krho, n=0):
...
+        return self.particle_kz * (1 - self._quad_pts)
+
     def _check_parameters(self):
-        must_be_specified = ['particle_kz', 'index_ratio',
-                             'size_parameter', 'lens_angle']
-        if any([getattr(self, p) is None for p in must_be_specified]):
-            raise ValueError("{} must be specified.".format(must_be_specified))
+        if any([getattr(self, p) is None for p in self.must_be_specified]):
+            msg = "{} must be specified.".format(self.must_be_specified)
+            raise ValueError(msg)
```
barkls, line 273: HoloPy actually has a `MissingParameter` exception class you might want to use here.

barkls (commented):
Nice improvements to code structure and thorough testing! I just had a couple minor comments that you should feel free to ignore.

This should be added to the [Scattering Theories User Guide](https://holopy.readthedocs.io/en/master/users/theories.html) that you wrote, but we can hold off until it's published if you prefer.

holopy/scattering/theory/mielensfunctions.py
briandleahy: Fixed by 08b8338d

briandleahy:
Your comments are addressed in the latest commits. I added a description of the AberratedMieLens theory in fedde9a0. I would appreciate it if you could take a quick look at that and tell me if my documentation is intelligible. Also, in the process I noticed that there were some out-of-date docs; I fixed these in baedc1ff. 

Let me know if you have other comments!

barkls:
Good catch on updating the function names in the dev guide! Everything else looks good except the posision typo

docs/source/users/theories.rst
```diff
@@ -96,6 +96,15 @@ Lens-Based Scattering Theories
       effects of a perfect objective lens.
     * Used for performance; ``MieLens(lens_angle)`` is much faster than calling
       ``Lens(lens_angle, Mie())`` and slightly faster than ``Mie()``.
+- :class:`.AberratedMieLens`
+    * Can handle :class:`.Sphere` objects, or :class:`.Spheres` through
+      superposision.
```
barkls, line 101: Typo

barkls (approved):

PR #387, opened 2021-03-13
Fittablescatteringtheory

### LINKED ISSUE #348: PerfectLensModel doesn't allow alpha fitting
The model available for Mielens calculations has the scaling parameter hard coded as 1 (calc_holo(detector, scatterer, theory=theory, scaling=1.0, **optics_kwargs) in PerfectLensModel class). This makes the version of mielens that's compatible with MCMC or cma-es fitting through holopy inconsistent with the paper's version of mielens, which does include alpha as a fitting parameter.

briandleahy:
Fixes #348 .

@barkls , before taking the time to do an in-depth code review, can you take a quick look at the overall changes I made and lmk if you think they are sufficient? Briefly
* `holopy.inference.model` now has a `theory_from_parameters` which gets called. Part of constructing this involved revamping the `Model._dummy_scatterer` attribute.
* `holopy.scattering.theory.ScatteringTheory` gets a `from_parameters` method. As a result, `ScatteringTheory` now has 2 new class attributes, a `parameter_names` and `unfittable_attrs`, which contain information about which parameters the `ScatteringTheory.from_parameters` needs to know about.

If so, I'll update the docs, both in "whats new" and the scattering tutorial.


barkls:
This looks good at first glance! The only thing I'm wondering about is whether there's redundancy in terms of defining a `ScatteringTheory` - it looks like any `__init__` parameters need to either end up in `parameter_names` or `unfittable attributes`. Is that right? If these two tuples are mutually exclusive and exhaustive of all `__init__` parameters you might be able to get rid of one of them by using the `_dict` property defined in `holopy_object.py`. You might even be able to simplify `from_paremeters` to:

```python
kwargs = self._dict
kwargs.update(parameters)
return self.__class__(**kwargs)
```
Untested code, might need a copy() in there to avoid modifying the theory.

Failing travis is due to a new xarray deprecation warning that we need to fix, but unrelated to this PR

briandleahy:
Good call! The `_dict` is a collection of the init args based on their value in the current object, which is what I want. There is some possibility of this causing a problem if (i) someone writes a class where the init takes arguments that aren't set as attributes, or (ii) we move away from yaml and the `_dict` property gets moved, but since `_dict` is used in a few places throughout I don't think it will be a problem. 

I'll add this in with some tests then make this a real PR.

barkls:
One more thing I noticed but required a bit more thinking regards what is stored in `model.theory`. For scatterers, we don't actually store the scatterer - we store a dummy scatterer that maintains the correct `from_parameters` and other attributes like `warn=True/False` for `Spheres`. We also store instructions for how to convert raw parameters into something compatible with the dummy scatterer's `from_parameters` in the maps dictionary. This organization ensures that we can easily store the model and all of its attributes as yaml regardless of what the initial scatterer looked like.

Right now, you aren't doing something similar for the theory. This could cause problems for cases where the theory object can't be written to yaml, prevent saving of the model or inference result object to a file. I believe this would only happen if the theory contains xarrays, e.g. if you want to specify a different lens angle (fixed value or prior) for two different illuminations, as defined by wavelength and polarization. Does the scattering theory even support this use case? If the answer is no, we're all good. If the answer is yes, we should either explicitly forbid this edge case, or work around it.

briandleahy:
A few thoughts:
1. I don't think the ScatteringTheory supports an xarray use case.... (Your multicolor code is pretty powerful, so I'm not 100% sure.) 
2. The reason I create the theory is because the default string `'auto'` obviously doesn't have a `.from_parameters` method. However, I could do something like what you did with the scatterer, like so:
```python
# in the init
self._dummy_theory = interpret_theory(self._dummy_scatterer, theory)
```
and
```python
    def theory_from_parameters(self, pars):
        pars = self.ensure_parameters_are_listlike(pars)
        formatted = read_map(self._maps['theory'], pars)
        return self._dummy_theory.from_parameters(formatted)
```
3. I'm a little confused why that works for the scatterer, since the `_dict` property should miss the scatterer then.....

barkls:
I don't think the xarray use case would work either, but it might be worth trying it. You're right theory needs to be handled differently to support 'auto' case. The current scatterer implementation is ~~pretty tricky!~~ actually broken, which I rubber duck'ed to realization while writing out how it works. The `Model.scatterer` property recreates the original scatterer that is passed in, which causes problems when loading from yaml if it included xarrays. I can't remember if there was another reason to create the `_dummy_scatterer` but it's failing at this purpose...

All that to say, I don't think you need to worry about this edge case for the theories. Let  me know if the PR is ready for a full review.

briandleahy:
Yeah, I think this is part of a bigger problem with yaml + xarray, so I'll leave it out of this PR. Some preliminary work towards it is on my branch yamlfittablescatteringtheory, in 1591c1b2 on my fork. I'll leave it up for a bit if someone wants to take a look at it.

I think this is ready for an actual review, if you'd like. @ralex0 I'd also appreciate your eyes on it if you have the time.

docs/source/releasenotes.rst
```diff
@@ -43,6 +45,11 @@ Developer Notes
   the scattering matrices or scattered fields needed to be calculated is
   now in `holopy.scattering.imageformation`.
 
+Deprecations
```
barkls, line 48: If you wanted to, you could use the FitWarning in holopy.fitting.py to temporarily maintain the PerfectLensModel namespace but also alert users that it won't stick around.

holopy/scattering/interface.py
```diff
@@ -134,7 +134,7 @@ def determine_default_theory_for(scatterer):
             theory = Mie()
     elif isinstance(scatterer, Spheroid) or isinstance(scatterer, Cylinder):
         theory = Tmatrix()
-    elif DDA().can_handle(scatterer):
+    elif DDA.can_handle(scatterer):
```
barkls, line 137: good catch!

holopy/scattering/tests/test_mielensfunctions.py
```diff
@@ -440,6 +441,26 @@ def test_raises_error_on_nans(self):
...
+            warnings.simplefilter('always')
+            s_theta = msm_highl._eval(theta)
+
+        self.assertFalse(np.isnan(s_theta))
+        # Then we check that this test actually worked, and that there
+        # was a nan warning raised:
+        messages = [str(r.message) for r in recorded_warnings]
+        should_be_warned = 'invalid value encountered in cdouble_scalars'
+        present = [should_be_warned in message for message in messages]
+        self.assertTrue(any(present), msg=f'This test is broken: {messages}')
```
barkls, line 461: I feel like there should be a cleaner way to do this, but not sure what it would be.

holopy/scattering/theory/lens.py
```diff
@@ -20,6 +20,7 @@ class Lens(ScatteringTheory):
     effect of an objective lens.
     """
     desired_coordinate_system = 'cylindrical'
+    parameter_names = ('lens_angle',)  # FIXME test
```
barkls, line 23: FIXME can be removed?

barkls (commented):
Looks good. Won't merge with failing tests though, so you can pin xarray on travis or wait for a PR from me including that fix.

holopy/scattering/theory/lens.py
briandleahy: Yes, I meant to remove that in fcd85986. I will delete it now.

holopy/scattering/interface.py
briandleahy: Thanks! It helps avoid the DDA instantiation, which can cause crashes because of our suppressing of fortran output. 

The other "catch" here was that the `DDA.can_handle` used to return True always, whereas now it returns `isinstance(scatterer, Scatterer)`.

holopy/scattering/tests/test_mielensfunctions.py
briandleahy: unittest.TestCase.assertWarnsRegex, which I've now used instead. Thanks!

briandleahy: (cf 7514e479)

holopy/scattering/theory/lens.py
briandleahy: (cf 7514e479)

docs/source/releasenotes.rst
briandleahy: The api is different enough, and the PerfectLensModel was around for short enough, where I'd rather not. 

If you want though, I can do something like

```python
class PerfectLensModel(object):
    def __new__(cls, *args, **kwargs):
        raise NotImplementedError("PerfectLensModel is deprecated; use an AlphaModel with a parameterized lens_angle instead")
```

Or I can leave it as-is (it is only 1 release, and I don't think people used the PerfectLensModel, although maybe I'm wrong.)

briandleahy:
> Looks good. Won't merge with failing tests though, so you can pin xarray on travis or wait for a PR from me including that fix.

Pinned travis to <0.17, which seems to be the breaking change (released in Feb). LMK what your thoughts are on the deprecating of PerfectLensModel and I'm ready to merge.

barkls (approved):

Issue #380, opened 2020-12-14
Scattering matrices are returned in a confusing format


briandleahy:
Currently, `ImageFormation.calculate_scattering_matrix` returns the scattering matrices in a confusing format, cf Line 160ff of `imageformation.py`:
```python
        coords.update({
            'r': (point_or_flat, r_theta_phi[0]),
            'theta': (point_or_flat, r_theta_phi[1]),
            'phi': (point_or_flat, r_theta_phi[2]),
            'Epar': ['S2', 'S3'],
            'Eperp': ['S4', 'S1'],
            })
```
The dimensions of the returned xarray should describe the two dimensions of the 2x2 scattering matrix, not just the two rows.

We should decide what is the best way to return the scattering matrices as an xarray.
(I would also advocate for a better naming convention than S1, S2, S3, and S4, since I can never remember how these relate to the parallel or perpendicular components....)

Since this is a user-facing change we should make sure that any change does not break existing workflows.

@barkls feel free to edit this issue as you see fit. (I'm not 100% sure I understand the problems you mentioned in the discussion on #377.)

vnmanoharan:
Re: the naming of S1-S4.  Our general (though unwritten) policy is to retain the notation used by Bohren and Huffman.  The notation here follows B&H, though the convention precedes them.

jeromefung:
FWIW, van de Hulst also uses the S1-S4 notation for the elements of the
amplitude scattering matrix, as do other codes like scsmfo1b and mstm. I
agree that it's not the most physically illuminating, but it seems to be
enough of a convention in the scattering literature that I would personally
argue against changing it.

Best,
Jerome

barkls:
Thanks for the feedback! Just to clarify, the main point of this PR isn't about the S1-S4 notation, but about the way the xarray dims and coords are defined. Right now, the matrix is defined like this:

![current naming](https://user-images.githubusercontent.com/19350183/102549767-ac945980-408a-11eb-8d3b-5218ce7eaa48.png)

The labels were presumably chosen to describe the rows of the matrix, but that's not actually what they represent as dims and coords. Here's a naming scheme that more correctly describes the transformation being done by the scattering matrix:
![correct naming](https://user-images.githubusercontent.com/19350183/102550406-c84c2f80-408b-11eb-8c85-cb3143294286.png)

PR #390, opened 2021-03-19
Various small changes

barkls:
Accidentally pushed this to develop branch (I thought GitHub disallowed this, but clearly trusted it too much). I created a new branch before my changes for the purpose of this PR.

Haven't updated releasenotes - let me know if there's anything here that should be included.

### DIFF
--- a/holopy/core/tests/test_io.py
+++ b/holopy/core/tests/test_io.py
@@ -230,6 +230,22 @@ class test_custom_yaml_output(unittest.TestCase):
         # only a & c appear because b is None, d is not in __init__ signature
         assert yaml.dump(instantiated) == '!S\na: 1\nc: 3\n'
 
+    @attr("fast")
+    def test_custom_treatment_of_ndarrays(self):
+        zero_length_array = np.array(3)
+        one_length_array = np.array([3])
+        longer_array = 3 * np.ones(10, dtype=int)
+        multi_D_array = 3 * np.ones((2, 2), dtype=int)
+        self.assertEqual(yaml.dump(zero_length_array), yaml.dump(3))
+        self.assertEqual(yaml.dump(one_length_array), yaml.dump([3]))
+        self.assertEqual(yaml.dump(longer_array), yaml.dump([3] * 10))
+        self.assertEqual(yaml.dump(multi_D_array), yaml.dump([[3, 3], [3, 3]]))
+
+    @attr("fast")
+    def test_custom_treatment_of_numpy_ufuncs(self):
+        yaml_sqrt = "!ufunc \'sqrt\'\n"
+        self.assertEqual(yaml.dump(np.sqrt), yaml_sqrt)
+        self.assertEqual(yaml.load(yaml_sqrt, Loader=yaml.FullLoader), np.sqrt)
 
 class TestMemoryUsage(unittest.TestCase):
     @unittest.skipIf(not importlib.util.find_spec('memory_profiler'),
--- a/holopy/core/tests/test_io.py
+++ b/holopy/core/tests/test_io.py
@@ -231,21 +231,34 @@ class test_custom_yaml_output(unittest.TestCase):
         assert yaml.dump(instantiated) == '!S\na: 1\nc: 3\n'
 
     @attr("fast")
-    def test_custom_treatment_of_ndarrays(self):
+    def test_custom_treatment_of_length_zero_ndarray(self):
         zero_length_array = np.array(3)
-        one_length_array = np.array([3])
-        longer_array = 3 * np.ones(10, dtype=int)
-        multi_D_array = 3 * np.ones((2, 2), dtype=int)
         self.assertEqual(yaml.dump(zero_length_array), yaml.dump(3))
+
+    @attr("fast")
+    def test_custom_treatment_of_length_one_ndarray(self):
+        one_length_array = np.array([3])
         self.assertEqual(yaml.dump(one_length_array), yaml.dump([3]))
+
+    @attr("fast")
+    def test_custom_treatment_of_longer_ndarray(self):
+        longer_array = 3 * np.ones(10, dtype=int)
         self.assertEqual(yaml.dump(longer_array), yaml.dump([3] * 10))
+
+    @attr("fast")
+    def test_custom_treatment_of_multidim_ndarray(self):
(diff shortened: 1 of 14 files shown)


briandleahy:
Hi @barkls  -- what do you want me to do here since these are already merged into develop? Do you want me to just review the changes and make suggestions, then not merge it in?

Alternatively, we could clean the history if we want to keep the dev branch "clean" always, but I don't think it's worth it given how small this is.

holopy/core/tests/test_io.py
```diff
@@ -230,6 +230,22 @@ def __init__(self, a, b, c=3):
         # only a & c appear because b is None, d is not in __init__ signature
         assert yaml.dump(instantiated) == '!S\na: 1\nc: 3\n'
 
+    @attr("fast")
+    def test_custom_treatment_of_ndarrays(self):
+        zero_length_array = np.array(3)
+        one_length_array = np.array([3])
+        longer_array = 3 * np.ones(10, dtype=int)
```
briandleahy, line 237: it might be worth splitting this up into 4 separate tests, that way if some change down the line in xarray or yaml does break this test, it will be more obvious which sub-cases need to be fixed.

holopy/core/tests/test_io.py
```diff
@@ -230,6 +230,22 @@ def __init__(self, a, b, c=3):
...
+        multi_D_array = 3 * np.ones((2, 2), dtype=int)
+        self.assertEqual(yaml.dump(zero_length_array), yaml.dump(3))
+        self.assertEqual(yaml.dump(one_length_array), yaml.dump([3]))
+        self.assertEqual(yaml.dump(longer_array), yaml.dump([3] * 10))
+        self.assertEqual(yaml.dump(multi_D_array), yaml.dump([[3, 3], [3, 3]]))
+
+    @attr("fast")
+    def test_custom_treatment_of_numpy_ufuncs(self):
```
briandleahy, line 245: likewise, it might be worth splitting this up into a "test_load" and test_dump", or alternatively just testing the round trip.

These are just suggestions, I don't think it's vital.

holopy/inference/result.py
```diff
@@ -273,10 +273,7 @@ def make_uncertain_value(p):
     def burn_in(self, sample_number):
 
         def cut_start(array):
-            if len(array.chain.coords) == 0:
-                array['chain'] = ('chain', array.chain)
-                array.set_index(chain='chain')
-            return array.sel(chain=slice(sample_number, None))
+            return array.isel(chain=slice(sample_number, None))
```
briandleahy, line 276: :+1:

holopy/scattering/imageformation.py
```diff
@@ -154,15 +154,15 @@ def _pack_scattering_matrix_into_xarray(
...
+        dims = [point_or_flat, 'E_out', 'E_in']
 
         coords = {point_or_flat: flattened_schema.coords[point_or_flat]}
         coords.update({
             'r': (point_or_flat, r_theta_phi[0]),
             'theta': (point_or_flat, r_theta_phi[1]),
             'phi': (point_or_flat, r_theta_phi[2]),
-            'Epar': ['S2', 'S3'],
-            'Eperp': ['S4', 'S1'],
+            'E_out': ['parallel', 'perpendicular'],
```
briandleahy, line 164: I much prefer this new naming convention, especially because (i) it is clear and (ii) these are the actual dimensions (e.g. in=parallel, out=perpendicular is correct and readable, not in=S1, out=S3 (say) which is both incorrect and not readable).

But I'm not sure if others want the S1, S2, S3, S4 etc convention explicitly stated somewhere, see the discussion in #380.

holopy/scattering/interface.py
```diff
@@ -102,8 +102,9 @@ def validate_scatterer(scatterer):
     # hierarchical data structures, so we check yaml serialization
     for prior_type in PRIOR_TYPES:
         if '!' + prior_type in yaml_scatterer:
-            msg = ('Scattering calculations require fixed values'
-                   ' but scatterer contains {} prior.'.format(prior_type))
+            msg = ('Scattering calculations require fixed values but scatterer'
+                   ' contains {} prior. Wrap Scatterer in a HoloPy Model and '
+                   'call model.initial_guess_scatterer').format(prior_type)
```
briandleahy, line 107: :+1: on the improved error message.

holopy/scattering/theory/mielens.py
```diff
@@ -148,6 +148,29 @@ class AberratedMieLens(MieLens):
 
     def __init__(self, spherical_aberration=0.0, lens_angle=1.0,
                  calculator_accuracy_kwargs={}):
+        """
```
briandleahy, line 151: woops, thanks for fixing my missing documentation!

briandleahy (commented):
Thanks for these bugfixes! The code looks good to me! 

The only obvious issue to me is the failing test on travis. Other than that, everything looks good. It would be nice if you could placate the bot's pep8 complaints. I also left 2 minor suggestions on tests. 

Regarding merging this in -- I would say that, since these commits are already merged in to develop, just fix any issues on your machine, run the tests locally and make sure they pass, then push straight to develop, Then I will close this PR once things are merged in. While that does risk pushing broken code to develop, I think it's OK since (i) you don't need to make a lot of changes and (ii) there already is a failing test on develop, and I think fixing that asap is a bigger priority. 

If you prefer to merge this in to develop the correct way, instead of pushing directly to develop, I can do that too as well.

holopy/core/tests/test_io.py
barkls: done

holopy/scattering/imageformation.py
barkls: I put a comment in the code. I could also refer to GitHub issue #380 since there's a nice diagram there that makes everything clear. What do you think?

barkls:
I think the best thing would be for you to actually merge this PR in (`develop` branch -> `dev` branch) when you are satisfied. Then we can pull `dev` back into `develop` although there won't be any changes. That way we get the merge in the commit history and this PR shows as merged rather than closed on GitHub.

I'm happy to use another workflow though if you prefer something else!

holopy/scattering/imageformation.py
briandleahy: I think this is fine.

Issue #322, opened 2020-03-06
Limits in order of magnitude?


MatthiasGi:
Hello there! I’m currently trying to use HoloPy as a starting point for my projects which happen to work on different orders of magnitude as the ones described in your publications or examples. To get to know the library better I tried to rebuild an experiment from the real world. The setup consists of spheres that are separated two millimeters each. A screen is placed 30cm away from the object. We use a red laser beam to achieve the hologram on the screen. The spheres itself are virtual and do not have a defined refractive index. They could be assumed as rock solid.

Unfortunately this doesn’t work:

```python
import holopy as hp
from holopy.scattering import Sphere
from holopy.scattering import Spheres
from holopy.scattering import calc_holo
import matplotlib.pyplot as plt
import scipy.constants as const

# Spheres
xOffset = 3 * const.milli
z = 300 * const.milli
params = {
    'r': 0.5 * const.milli,
    'n': 100
}
s1 = Sphere(center = (xOffset, 0, z), **params)
s2 = Sphere(center = (xOffset + 2 * const.milli, 0, z), **params)
s3 = Sphere(center = (xOffset, 2 * const.milli, z), **params)
collection = Spheres([s1, s2, s3])

# Simulation
medium_index = 1
illum_wavelen = 632 * const.nano
illum_polarization = (1, 0)
resolution = 2500
detector = hp.detector_grid(shape = resolution, spacing = 5 * const.centi / resolution)

# Holography
holo = calc_holo(detector, collection, medium_index, illum_wavelen, illum_polarization)

# Output
hp.show(holo)
plt.show()
```

The produced error is `holopy.scattering.errors.InvalidScatterer: Invalid scatterer of type Sphere.
radius too large, field calculation would take forever`. Which is understandable as `k_medium * particle_radius ≤ 1e3` with `k_medium = 2π/lambda_0 * n_medium` seems to be the given limit in the implementation of Mie-Theory. Even reducing the particle size to e.g. 1µm doesn’t solve the problem, as now the separation is too large.

This seems to be a limitation in the theory used to calculate the given problem. I already tried to submerge myself in DDA – but: there too seem to be limitations as the calculation takes so long that it will be cancelled by my operating system. Before now investigating more: Is there maybe already another theory implemented which handles these cases? Can you maybe point me in the right direction?

Apart from that: Could it be helpful to clearly document the (current) limits of HoloPy?

Thanks in advance for your kind help or any tips!

vnmanoharan:
[...]

PR #393, opened 2021-04-02
Fix #338 and #322

### LINKED ISSUE #338: `determine_default_theory_for` picks Multisphere for two spheres even when they are far apart, which causes problems
See this discussion, take from @jeromefung in #337 
> This stemmed from a recent email exchange I had with a HoloPy user who was obtaining some unphysical results when calculating holograms with the default scattering theory (Multisphere) for multiple spheres separated by many particle diameters. There's nothing in the documentation to suggest using Mie in such a case.

> The real issue the user was encountering was that Multisphere expands the scattered fields in vector spherical harmonics whose origin is at the center of mass of all the spheres. Apparently, the default limits on the expansions by scsfmo1b weren't enough for the user's situation (but in that situation, optical coupling was negligible).

One solution would be for `determine_default_theory_for` to check if the spheres are far enough apart where Mie Superposition would be a better approximation than Multisphere. 

However, this solution could be problematic during an inference calculation if the spheres sit near the cutoff distance for switching between Mie and Multisphere. I can imagine a situation where the best fit with Multisphere places the spheres far enough apart for Mie Superposition to be used, and the best fit with Mie Superposition places the spheres close enough together to use Mie Superposition, and the inference algorithm constantly switches back and forth between those two. If someone is unlucky this could cause convergence issues unless the theory is specified once per inference calculation and does not switch. 

As a temporary measure we can add a discussion in the user guide regarding this.

briandleahy:
This PR makes `determine_default_theory_for` give Mie instead of Multisphere when the spheres are far apart. cf. #338 . There is a large in-line comment explaining the thought process behind the change; let me know if a different approach would be better.

In addition, I also updated the docs to discuss holopy's computational limitations for large particles a bit more. I can expand on that if needed.


holopy/scattering/interface.py
```diff
@@ -355,3 +351,47 @@ def scattered_field_to_hologram(scat, ref):
...
+        # self-consistently solve for multiple-sphere scattering.
+        # For computational reasons Multisphere does not go past a fixed order
+        # (=70), but the actual number of terms needed scales as kR.
+        # So for large R, the error in Multisphere scales as
+        #       error_multisphere = kR / 100
+        # where the factor of 100 is the value of kR at which
+        # multisphere stops increasing the expansion order.
+        # The error for multisphere is smaller than the Mie
+        # superposition error when
+        #       (a / R)^2 / 100 < 1
```
barkls, line 388: I think this might be backwards? This statement has multisphere error being smaller than Mie for very large R but I think that's the opposite of what we want.

holopy/scattering/tests/test_multisphere.py
```diff
@@ -155,36 +155,55 @@ def test_radial_holos():
...
+            r=5e-07)
+        sphere2 = Sphere(
+            center=[6e-6, 5e-6, 10e-6],
+            n=[1.5, 1.6],
+            r=[2e-7, 5e-07])
+        scatterer = Spheres(scatterers=[sphere1, sphere2])
+        self.assertRaises(
+            TheoryNotCompatibleError,
+            calc_holo,
+            schema, scatterer, *self.args, **self.kwargs)
```
barkls, line 206: Thanks for reformatting these tests!

barkls (approved):

holopy/scattering/interface.py
```diff
@@ -355,3 +351,47 @@ def scattered_field_to_hologram(scat, ref):
...
+        # Note that the Mie error could still be large, if ka >> R / a
+        max_radius = max([sphere.r for sphere in spheres.scatterers])
+        centers = np.array([sphere.center for sphere in spheres.scatterers])
+        dx = centers.reshape(1, -1, 3) - centers.reshape(-1, 1, 3)
+        max_separation = np.linalg.norm(dx, axis=2).max()
+        close_enough = max_separation <= 15 * max_radius
```
barkls, line 394: Where did the 15 come from? You might want to replace it with np.sqrt(100) to make clear the correspondance with the expression in comments above.

holopy/scattering/interface.py
briandleahy: Sorry, I'm bad at math. I clarified the confused derivation I had there; it should be fixed in e5cf24e5

holopy/scattering/interface.py
briandleahy: (also fixed in e5cf24e5)

barkls:
closing and reopening to trigger travis build

PR #405, opened 2021-11-03
Do not use space in directive.

Carreau:
This is non standard and not supported by all rst parsers

### DIFF
--- a/holopy/core/errors.py
+++ b/holopy/core/errors.py
@@ -19,7 +19,7 @@
 """
 Error classes used in holopy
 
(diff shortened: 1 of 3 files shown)


briandleahy:
Thanks for the fix @Carreau !

@barkls , any idea why the doctest is failling?

Carreau:
> @barkls , any idea why the doctest is failling?

There is a new recent version of docutils that apparently breaks everything.

Carreau:
And yes, the error is described in this new RDT blog post: 

https://blog.readthedocs.com/build-errors-docutils-0-18/

barkls (approved):

Issue #430, opened 2024-01-03
Calculation of noise_sd should be revised


vnmanoharan:
In [`holopy.core.io`](https://github.com/manoharan-lab/holopy/blob/3bed86e3b493245deb9d8d74904c991bf13caaf5/holopy/core/io/io.py#L552), we estimate the pixel noise from a stack of images as follows. We first calculate the standard deviation for each pixel in each channel over the stack, then calculate the coefficient of variation (CV) for each pixel in each channel, then average the CV over all pixels for each channel. We end up with an estimate of the pixel noise in each channel. Let $`I_{ijkl}`$ be the stack of images, where $`i`$ and $`j`$ are the 2D coordinates of each image, $`k`$ is the channel, and $`l`$ is the image number in the stack.  The pixel noise calculation we currently do is
```math
\mathrm{noise\_sd}_k = \mu_{ij} \left(\frac{\sigma_{l}( I_{ijkl} )}{\mu_{l} (I_{ijkl})}\right),
```
where $`\sigma_l`$ is the standard deviation calculated over the stack, $`\mu_l`$ is the mean calculated over the stack, and $`\mu_{ij}`$ is the mean calculated over image coordinates.

There are other ways to calculate the pixel noise. An alternative approach is to take the pixel-wise mean of the numerator and denominator above, and *then* take the ratio:
```math
\mathrm{noise\_sd}_k = \left(\frac{\mu_{ij} \left[\sigma_{l}( I_{ijkl} )\right]}{\mu_{ij} \left[\mu_{l} (I_{ijkl})\right]}\right).
```

We used to do the second method, until https://github.com/manoharan-lab/holopy/commit/89450e00f21cef98afd4e93605ffba13beef9679, when we changed to the current method.  The current estimator, though, has some undesirable properties. In particular, the estimate of the pixel noise will be dominated by pixels with high CV, and the CV can be infinite if the mean value of the pixel is zero. So in general the first estimator should give a higher noise level than the second, and it will be more sensitive to outliers.

We should probably go back to the second method, but since the estimated pixel noise is used in the likelihood function, changing the calculation now would change inference results. Targeting this change for version 4.

PR #423, opened 2024-01-02
Update code to work with recent versions of xarray, ensure tests pass

vnmanoharan:
As of `xarray` v2022.06.0, arrays with a `MultiIndex` no longer have associated "virtual" coordinates but instead have real coordinates (https://docs.xarray.dev/en/stable/whats-new.html#v2022-06-0rc0-9-june-2022) . Some of our code had assumed that the x, y, z dimensions associated with a flattened detector schema would remain virtual, but that code broke when they became non-virtual, and the x, y, z dimensions started getting duplicated in our metadata update operations. This PR fixes this problem, allowing `holopy` to be used with more recent versions of `xarray`.  

Fixes 
#409
#411
#415

Other changes:

- added environment.yml file to make it easier to setup a conda environment for developing holopy
- fixed some deprecation warnings associated with changes in `scipy.ndimage`
- created workaround for incompatibility of `nose` with `unittest.expectedFailure`.

`test_2_colour_noise_sd (test_io.TestAccumulator)` is failing, but other tests are passing. This test was failing before any changes were made.

PR #431, opened 2024-01-03
Re-gold two-color noise test

vnmanoharan:
[`test_2_colour_noise_sd`](https://github.com/manoharan-lab/holopy/blob/3bed86e3b493245deb9d8d74904c991bf13caaf5/holopy/core/tests/test_io.py#L329-L336) is failing for me. I have tried several approaches (using numpy on the raw images in addition to using the code currently in `holopy.core.io.io.py`) to reproduce the gold values of the noise_sd that are reported in the test, but I can't reproduce these values. This might be related to #430, in that a different method was used to generate the gold values than the one we currently use. In any case, it seems that changing the gold values is the simplest approach here.

This PR also cleans up some of the code in `load_average()` and the `Accumulator` class. In particular, I removed the method `accumulator.cv()` since it does pixel-wise averaging for multiple channels, and the `Accumulator` class doesn't require a labeled array and so is ignorant of coordinates and channels (as it should be). The calculation of the coefficient of variation is now in `load_average()`. I removed the tests associated with the `accumulator.cv()` method and added a new test to verify the numerical accuracy of the Welford algorithm implementation in `accumulator.push()` (see #276 for why we needed to implement this approach). The new code now takes advantage of `xarray` coordinate labels.

### DIFF
--- a/holopy/core/io/io.py
+++ b/holopy/core/io/io.py
@@ -478,11 +478,9 @@ def load_average(
         spacing = get_spacing(refimg)
 
     # read colour channels from refimg
-    channel_dict = {'0': 'red', '1': 'green', '2': 'blue'}
+    channel_dict = {'red': 0, 'green': 1, 'blue': 2}
     if channel is None and refimg is not None and illumination in refimg.dims:
-        channel = [
-            i for i, col in enumerate(['red', 'green', 'blue'])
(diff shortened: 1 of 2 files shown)


vnmanoharan:
commits for this PR moved to #423.

PR #433, opened 2024-01-06
Fix output suppression from multisphere and DDA codes

### LINKED ISSUE #403: SuppressOutput crash IPython (close stdout too many times).
See https://github.com/ipython/ipython/issues/13223

Smallest reproductible example:

```python
In [1]: from holopy.core.utils import SuppressOutput
   ...: with SuppressOutput():
   ...:     pass
```

My understanding is that in 

```python
class SuppressOutput():

    def _redirect_stdout(self, destination_fileno):
        if self.stdout_behaves_normally:
            sys.stdout.close(). # this is not yours to close, you haven't opened it.
        os.dup2(destination_fileno, self.std_out)
        if self.stdout_behaves_normally:
            sys.stdout = io.TextIOWrapper(os.fdopen(self.std_out, 'wb'))
```

It's ok to replace it, but you should restore it after the context manager. 
and in general you should not close an object you did not open as someone else might have references to ti.

Storing sys.stdout in enter, and restoring it in exit seem to work.

vnmanoharan:
The `SuppressOutput` context manager was wreaking havoc with `pytest`. The problem, as noted in #403, is that the context manager closes `stdout`, which `pytest` (and `ipython`) expect to remain open. We originally did this because the multisphere code keeps outputting to `stdout` even when python attempts to redirect `stdout` to `/dev/null`. This is a general problem when wrapping fortran codes in python. This PR fixes this issue by removing the context manager and modifying the `amncalc` subroutine in `scsmfo_min` to not print anything unless it is told otherwise. The other place that the `SuppressOutput` context manager was used was in the code that interfaces with DDA. Since we call ADDA in a subprocess, the output can be suppressed by using the `capture_output` kwarg of `subprocess.run()`.

Fixes #403. With these changes, `pytest` now runs, though there is some additional work to do to fully integrate it.

### DIFF
--- a/holopy/core/utils.py
+++ b/holopy/core/utils.py
@@ -41,41 +41,6 @@ except ModuleNotFoundError:
 from holopy.core.errors import DependencyMissing
 from holopy.core.holopy_object import HoloPyObject
 
-
-# FIXME this is difficult to test, since it works on the file level
-# perhaps make this capture the output rather than writing it to devnull?
-class SuppressOutput():
-    def __init__(self, suppress_output=True):
-        self.suppress_output = suppress_output
-        try:
-            self.std_out = sys.stdout.fileno()
-        except io.UnsupportedOperation:
-            self.stdout_behaves_normally = False  # ie running on travis
-            self.std_out = 1
-        else:
-            self.stdout_behaves_normally = True
-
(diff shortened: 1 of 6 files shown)


carolinesmartin (approved):
Looks good, pre existing code still runs

PR #436, opened 2024-01-10
fixed bug with detrend and xarray and added test

### LINKED ISSUE #410: Propagate/Detrend
Hi all,

First question: how do I use 'propagate' to estimate the z distance of a scatterer from a hologram?

Second: Detrend produces a value error when used on the example holograms. The other functions from the same submodule do not. How do I fix this issue?

Vitan

vnmanoharan:
Fixes #410


carolinesmartin (approved):
Looks good. Everything runs properly.

PR #438, opened 2024-01-11
fixed undefined symbols identified by flake8

vnmanoharan:
This PR fixes a lot of undefined variables, classes, and function names found by running
```
flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
```
These would have led to runtime errors but the undefined symbols were in places that aren't typically reached in normal execution.


carolinesmartin (approved):
Looks good. Everything runs properly.

